PR #6, opened 2016-07-14
Using a shared cache for R algorithms

jamesatha:
Let's just pretend like this never happened...


jamesatha:
No objections? sweet!

PR #11, opened 2017-02-09
upgrade rust template and client

anowell:
This does require updating the docker image,
  as the changes depend on Rust 1.15 in the builder image.

This has no effect on published algos, but does break
  compilation of existing rust algorithms.
  The impact is quite small (namely myself, @zeryx,
  and a small handful of unpublished, stale test algos)
  The work-around is to just update dependencies
  and switch from rustc-serialize to serde
  which is often just an attribute change except in
  cases of complex custom JSON parsing.

The improvements in this change include:
- Using the new 2.0 client
- Error messages should be more thorough ("caused by" chain)
- Adds algorithm testing (previously held back for ergonomic reasons)
- Switches to serde serialization/deserialization which is faster and significantly more flexible/configurable (and finally works on stable rust - I've been holding back this change for months, waiting on rust 1.15's stabilized support for serde)
- Algorithm boilerplate is simpler (about half as many lines of code, and hides the complex type signatures) thanks to `algo_entrypoint` macro.

### DIFF
--- a/rust/template/Cargo.toml
+++ b/rust/template/Cargo.toml
@@ -2,7 +2,7 @@
 name = "__ALGO__"
 version = "0.1.0"
 
-# Do NOT change the [[bin]] and [lib] sections
+# Changing [[bin]] and [lib] sections may have unintended consequences
 [[bin]]
 name = "pipe"
 path = "bin/pipe.rs"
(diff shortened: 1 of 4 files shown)


anowell:
note: this PR cannot be merged until we deploy an builder image in Rust 1.15 (no code changes required - just any builder image built since last Feb 2), otherwise it will be impossible to compile new rust algorithms.

So this is currently blocked by mxnet building issues...

PR #14, opened 2017-10-02
Compile speed

pmcq:
Pretty big re-configurating of image creation:
 - Adding scripts for installing all of the various libraries that we have
 - Add script for templating a Dockerfile for a particular set of libraries
 - Splitting out `python2` as a separate template from `python3` (especially to simplify testing of changes)
 - No longer have single `bin/build` but have two separate `bin/install-dependencies` and `bin/install-algorithm` which get run as individual steps so that we can cache installed deps without needing to re-install on every compile (really what most of this work has been all about)
 - No longer need `bin/setup` as part of the algorithm template and instead that belongs in the libraries directory


anowell:
Looks good on quick read. Could you also update the main `README.md` with relevant changes to the build process? (or is that all unchanged)

pmcq:
Currently it is an unchanged build process for all of the languages except for the special `python2` that has an installer in the `libraries` directory.  I can add a note in the readme for building the new special language.  After doing more integration testing with `python2` will port over the other languages as well and finally update the full readme.

PR #20, opened 2018-01-12
Remove driver from builder image

pmcq:
a version of cuda remains so that packages like pycuda can be installed via pip since they build against the cuda headers


jamesatha:
:+1: 
@zeryx or @besirkurtulmus want to merge this. It looks fine to me and am too afraid to hit the button :fearful:

PR #21, opened 2018-03-19
replacing relative URL with absolute

zeryx:
Some Algorithms, like https://algorithmia.com/algorithms/weka/WekaClassification use runtime reflection to load bundled jar files when necessary. With adock, this worked fine for our weka algorithms - however with the obsolescence of Adock and it's replacement by langpacks this has introduced a breaking bug to our wekaClassification algorithm and it's derivatives.
[ALGO-351](https://algorithmia.atlassian.net/browse/ALGO-351)
smallest reproduction: [https://algorithmia.com/algorithms/weka/weka_scala_test/edit](https://algorithmia.com/algorithms/weka/weka_scala_test/edit
)
From the weka package, we were able to narrow down the failing code block to this is here:
```
File file = new File(new URI(part));
```
where `part` defines a classPath URI recovered from `getClass()` being called on the packaged class.
When we are using relative class paths like `file:./lib_managed/some/package.jar` the weka code fails with the following message: `java.lang.IllegalArgumentException: URI is not hierarchical`.

We were able to narrow down where the classPaths are constructed for an algorithm to here, and we discovered that when a relative path (aka `new File(".")`) was passed into the JarRunner, this would replicate the bug we're experiencing.

Bear in mind that no experimental testing has been performed, besides basic "does this break any other part of the JarRunner code?" sanity checks. As I don't have exposure to our CI tools I will defer this to someone more experienced.

### DIFF
--- a/java/pipe/src/main/java/com/algorithmia/JarRunner.java
+++ b/java/pipe/src/main/java/com/algorithmia/JarRunner.java
@@ -200,8 +200,8 @@ public class JarRunner {
     private ClassLoader loadJars(String workingDirectory) {
         // Load JARs
         String[] extensions = {"jar"};
-        File absolutePath = new File((new File(workingDirectory)).getAbsolutePath());
-        Collection<File> jarFiles = FileUtils.listFiles(new File(absolutePath), extensions, true); // Recursively find jars
+        File absolutePath = (new File(workingDirectory)).getAbsoluteFile();
+        Collection<File> jarFiles = FileUtils.listFiles(absolutePath, extensions, true); // Recursively find jars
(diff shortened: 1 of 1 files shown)


jamesatha:
Looks good to me

zeryx:
Ideally there would be a way to test whether or not this actually does what I think it does - aka if I call `classPath()` in an algorithm's apply method that is loaded using the JarRunner, will the values here "stick"
I'm not sure how to do that without making a deploy. Nothing stopping us from loading it up in test, though.

neildf:
It shouldn't be that hard to write a test up to prove locally that it works as expected.

zeryx:
seems to work properly locally, merging

PR #27, opened 2018-07-27
Add a default error_type when deserializing

anowell:
@kennydaniel this should address [PLATFORM-1180](https://algorithmia.atlassian.net/browse/PLATFORM-1180).

"it compiles", so it's gotta work... ;-)

For context/rust-education, this uses [serde field attributes](https://serde.rs/field-attrs.html). It will handle the case where `error_type` is missing, but it will still fail to parse if given non-string values (including `null`). If we want to support fallback for `null` or for any non-string value, it's a bit more complicated (custom deserializer) but still only a few more lines of code: [example of alternatives](https://play.rust-lang.org/?gist=dc058ccb5b271de6f01619ab86544fe5&version=stable&mode=debug&edition=2015). I'm somewhat tempted to use the default `error_type` for any parsing error, but figured I'd throw out the simplest fix first for review.


kennydaniel:
This looks good to me. I don't actually know what the procedure is for testing a new langpack though.

anowell:
I manually tested this without a full stack doing the following:

```
cargo build

$ bin/make-example ruby  # cuz I know ruby well enough to debug a ruby example
$ cd /tmp/algorithm  # where the example is created
$ bundle install  #cuz ruby
$ ~/algo/langpacks/target/debug/langserver
> Listening on port 9999.

# different shell, just sending input that will raise an exception (ruby uses error_type: e.class.name)
$ curl localhost:9999 -X POST -H 'Content-Type: application/json' -d '[]'
> {"error":{"message":"no implicit conversion of Array into String","error_type":"TypeError",...

# remove setting the error_type from bin/pipe.rb. Fallback error_type is working...
$ curl localhost:9999 -X POST -H 'Content-Type: application/json' -d '[]'
> {"error":{"message":"no implicit conversion of Array into String","error_type":"AlgorithmError", ...

# test with error_type => nil. As predicted, null isn't supported with this PR
$ curl localhost:9999 -X POST -H 'Content-Type: application/json' -d '[]'                                                                             
> {"error":{"message":"JSON error - data did not match any variant of untagged enum RunnerOutput","error_type":"SystemExit" ...
```

Ideally we'd create some automatic test to go through this flow. But overall, I consider this adequate to merge, but could use someone to help shepherd this into the runners.

PR #28, opened 2018-08-02
Creating new languages directory for language-specific installation & dockerfiles

pmcq:
Using python2 as a testbed for new development work which involves new jinja templates for dockerfiles.  The old python/template/bin directory has roughly been copied to languages/python2/bin (with a couple tweaks to filenames) along with a new config file for the language installation.  The templated Dockerfile.compile in languages/python2 is not used except for testing purposes, but the templated Dockerfile.runner and .builder are used by jenkins for building the base images.  The bin/build-language script can be used to generate the templated dockerfiles in `languages/python2` as well.

Since this is still in a semi-experimentation phase it's possible some of these commits will later be backed out or tweaked but is worth merging to master as it is functional and other repos are moving to master as well.


languages/Dockerfile.compile.j2
```diff
@@ -0,0 +1,12 @@
...
+COPY --chown=algo algosource /opt/algorithm
+RUN /opt/algorithm/bin/build
+
+FROM {{runner_image}}
+{% for artifact in config.build.artifacts %}
+COPY --from=builder --chown=algo {{artifact.source}} {{artifact.destination}}
+{% endfor %}
+USER algo
+WORKDIR /opt/algorithm
+ENTRYPOINT /bin/init-langserver
```
jamesatha: Shouldn't this entrypoint be the build script? (the same is true in the other compile file)

languages/python2/bin/build
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+set -e
+
+if [[ -n "$PYTHON_INDEX_URL" ]]; then
```
jamesatha: Should there be an ARG in the docker file for this? Not totally sure how this would be set in the build pipeline

jamesatha (approved):
Looks alright to me

languages/python2/bin/build
pmcq: Great point! Added JIRA item to track setting environment variables for mirrors (and http proxy...? that might be a separate work item)

kennydaniel:
Is this moving us away from anaconda and toward using system python 2/3?

pmcq:
Nope, the python2 install is still using the same version of anaconda python: https://github.com/algorithmiaio/langpacks/pull/28/files#diff-b2f581c9a946a13d5b749e923b35a9dc but there's just two separate dockerfiles, and the install is way cleaner/lighter (removed duplicated packages etc).  We could create a separate langpack w/o anaconda if desired though.

kennydaniel (approved):

PR #30, opened 2018-08-09
Add parameters so that the base image and a suffix can be templated into the dockerfiles

pmcq:


pmcq:
I mostly liked having a fully templated file that we could check in to source control or ship elsewhere if we needed.  I also just feel like I've been annoyed by missing a `--build-arg` flag in the past and was trying to avoid that haha.

PR #32, opened 2018-08-14
rstrip searches for characters not strings :facepalm:

pmcq:


peckjon (approved):
Modulo the small chance that there’s a package with .tar.gz validly in the middle — eff that — LGTM I guess?

kennydaniel (approved):

kennydaniel:
@peckjon I thought the same thing... so opened a python console and tested if `replace('.tar.gz$', '')` worked, but it apparently isn't a regex. And then I said the same as you... eff that, what are the chances of an R package named uncompress.tar.gz.tar.gz :-)

PR #37, opened 2018-12-18
Big refactor so that python language is installed as a library instead of as a separate entity

pmcq:
This changes the contract a bit for:
a) what is needed in order to be a valid language in the platform
b) how images are based for an algorithm
c) how we would update the version of langserver in use

At a high-level:
1. The langserver binary will be built in an image and published by itself when updates occur
2. When runner images are built the langserver binary will be copied in from some published langserver image (this makes it easier to use anything arbitrary as a base image as long as langserver is a static binary.  we might need to copy other files in the future as needed)
3. Language build/runtimes would be installed no differently than any other system package/library (and the build vs runtime will be represented as two distinct libraries)


adnaanm:
Should `langpacks/languages/python2/Dockerfile.builder`  still references `languages/python2/bin/install-buildtools` even though it looks like that file doesn't exist anymore? Looks like you renamed/moved that stuff to the `libraries` folder. Am I missing something?

adnaanm:
And in the same file, the lines

```
COPY languages/python2/bin/build /opt/algorithm/bin/build
COPY languages/python2/bin/test /opt/algorithm/bin/test
```

look like they should be targeting `libraries/python2-buildtime/context/(build | test)` I think?

languages/python2/config.json
```diff
@@ -1,21 +1,7 @@
...
-            {"name":"PATH", "value":"/opt/anaconda2/bin:$PATH"}
-        ],
-        "files": [
-            {"source":"bin/pipe", "destination":"/opt/algorithm/bin/pipe"}
-        ]
-    }
+    "artifacts": [
+        {"source":"/home/algo/.local", "destination":"/home/algo/.local/"},
+        {"source":"/opt/algorithm/src", "destination":"/opt/algorithm/src/"},
+        {"source":"/opt/algorithm/algorithmia.conf", "destination":"/opt/algorithm/algorithmia.conf"}
```
jamesatha: From a debug-ability standpoint, it would be nice to also copy over the requirements file

Also, it seems like templates might have other dirs that are siblings to src. Can we just take all of /opt/algorithm?

libraries/python2-buildtime/install.sh
```diff
@@ -9,3 +9,5 @@ apt-get install -u -y \
     python-pip
 
 pip install --upgrade pip
+
+rm -rf /var/lib/apt/lists/*
```
jamesatha: nit: ideally, don't you want to delete the files that have a create timestamp greater than when we started running this script?

languages/Dockerfile.builder.j2
```diff
@@ -3,25 +3,18 @@ FROM {{base_image}}
 # Set options that should be defined everywhere
 ENV JAVA_TOOL_OPTIONS=-Dfile.encoding=UTF8
 ENV LANG C.UTF-8
+ENV LANGPACKS_VERSION={{langpacks_version}}
```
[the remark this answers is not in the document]
pmcq: because i felt there was otherwise no way to tell what version of langpacks a given image was built from (something that can be annoying now as well).  if you exec into a built image - what version of the template files are in there?

languages/Dockerfile.runner.j2
```diff
@@ -1,34 +1,22 @@
...
-COPY src /langserver/src
-WORKDIR /langserver
-RUN cargo build --release
-
 FROM {{base_image}}
 
 # Set options that should be defined everywhere
 ENV JAVA_TOOL_OPTIONS=-Dfile.encoding=UTF8
 ENV LANG C.UTF-8
+ENV LANGPACKS_VERSION={{langpacks_version}}
```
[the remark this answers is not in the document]
pmcq: sim to above - and i didn't think the `FROM` statement is preserved in the built images but i'm not positive on that... open to other suggestions on how to preserve that metadata

libraries/python2-buildtime/install.sh
pmcq: do you mean just within the `/var/lib/apt` directory? it starts out as empty in the base ubuntu image (and most others i've tested with)

pmcq:
> In `langpacks/languages/python2/Dockerfile.builder` the COPY command still references `languages/python2/bin/install-buildtools` even though it looks like that file doesn't exist anymore? Looks like you renamed/moved that stuff to the `libraries` folder. Am I missing something?

yeah thanks - we actually don't need those files at all any longer (we only need the template files).  if i had re-run the build-language script then those files would have popped out correctly, but I don't see a reason to have them checked in and so just removed them.

languages/python2/config.json
pmcq: hmm... I guess I hadnt looked at any python templates yet (and seems weird  they'd have things other than src but idk). I think copying the entire directory might have clobbered the pipe script that is already in the runner image... maybe if the destination has a trailing / or something it would work but it's been like 5 months since this was made so I cant recall.

jamesatha: yeah, that would clobber the bin/pipe script - I do think the requirements file would be nice to have. 

I feel like this is fine for now and am not sure if we want to think about a way to let people customize it in the future

libraries/python2-buildtime/install.sh
jamesatha: Yeah, just talking about /var/lib/apt/lists/. I am guessing that you did this to keep the images small. And I wanted to make sure that we weren't adding layers for the removal of files from the base. If it's empty, this lgtm

languages/python2/config.json
neildf: Arguably all of the stuff living at the root of templates should be published as packages. 

That said, doesn't change the fact that there might be other things at the root that should be copied

languages/Dockerfile.runner.j2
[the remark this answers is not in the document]
neildf: (And labels are preserved from parent images but can be overwritten by the current image)

libraries/python2-runtime/install.sh
```diff
@@ -9,3 +9,5 @@ apt-get install -u -y \
     python-pip
```
pmcq: talking with adnaan yesterday realized we probably don't need pip in the runtime image at all? anyone else have thoughts on that?

languages/python2/config.json
pmcq: i added in the requirements.txt - think that's a good idea.  i'm not sure if we'll need to customize it in the future, but when we actually have templates to  look at we can reconsider this.  alternatively, the templates might have a config file indicating all the things to copy too? Idk

libraries/python2-runtime/install.sh
jamesatha: Sounds good to me

jamesatha: Also, kind of random but it seems like some versions (or more recent versions?) of apt-get don't have the --show-upgraded (-u) flag. My machine doesn't have it but I can see ubuntu 16.04 does

pmcq: Actually realized we need it to install packages into the runtime image (e.g. so tensorflow is installed)

libraries/python2-buildtime/Dockerfile
```diff
@@ -0,0 +1,2 @@
+COPY python2-buildtime/build /opt/algorithm/bin/build
+COPY python2-buildtime/test /opt/algorithm/bin/test
```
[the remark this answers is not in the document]
pmcq: hmm... that's a good point - i had manually tested the dockerfile/context and I might have set the path incorrectly... i'll verify this now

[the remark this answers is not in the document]
pmcq: thanks for pointing this out - just updated to include `context` in there too

PR #38, opened 2018-12-18
ALGO-526 creating a static keras/tensorflow algorithm template

zeryx:
as mentioned here:
https://algorithmia.atlassian.net/browse/ALGO-526
We're trying to create some static package-set templates for a few ml frameworks, this is a proposed PR for the keras/tensorflow one.


templates/tensorflow-gpu-1.12/image_classifier/src/example_test.py
```diff
@@ -0,0 +1,22 @@
+from .example import apply, get_model
+import Algorithmia
+
+"""
+This is the testing file, which should be bundled along with the Algorithm.py file when creating a template.
+These tests are run during the creation of a template, and if they fail - the templating process fails as well.
```
pmcq: we won't be running any tests while creating templates for the current iteration - i'm not sure if this file should be removed or not

templates/tensorflow-gpu-1.12/image_classifier/src/example.py
```diff
@@ -0,0 +1,57 @@
+import Algorithmia
```
pmcq: i think this file should be named __ALGORITHM__.py (which will get autofilled in with the name of the algorithm).  the pipe script is going to try loading a file with that name, and otherwise an algorithm made with this template won't have a main entrypoint

templates/tensorflow-gpu-1.12/image_classifier/src/helpers/download_helpers.py
```diff
@@ -0,0 +1,100 @@
...
+	else:
+		algo_input = image_url
+	img_data_path = client.algo("util/SmartImageDownloader").pipe(algo_input).result["savePath"][0]
+	return client.file(img_data_path).getFile().name
+
+def load_pickle_file(pickle_file_path, local_file = True, client=None):
+	"""
+	Download and load a pickle file into a python object. The file can exist in Algorithmia (data://) or any public url or locally
+	Parameters:
+	location: Absolute path to the file (remote or local)
```
pmcq: not going to judge any of the actual python code much but i think there's conventions out there for specifying parameter names in docstrings that numpy uses that other tools work well with.  something like `:param location str: blah blah` https://stackoverflow.com/a/40596167

templates/tensorflow-gpu-1.12/image_classifier/src/helpers/download_helpers.py
```diff
@@ -0,0 +1,100 @@
...
+	Description:
+		checks the type of a field in a dictionary, if it's not one of the defined types, it throws an exception.
+	Parameters:
+	dic: The dictionary object that we should check against.
+	id:  The specimen field name.
+	typedef: The type or list of types that your field type should be a member of.
+	Returns:
+		The object contained in the dictionary at the objective field.
+	"""
+	if isinstance(typedef, type):
```
pmcq: also this feels a little weird cause python ain't about types usually but feels like having this be `type_check(dic, id, *typedef)` and that way you remove at least one of the code paths might be nice.  again :shrug: not really my place to judge on this

templates/tensorflow-gpu-1.12/image_classifier/src/example_test.py
zeryx: I kinda want to have a x_test.py file there that actually does something, but I can remove it / remove the comment

templates/tensorflow-gpu-1.12/image_classifier/src/example.py
zeryx: oh ok I wasn't sure what the pipeline looked like, I'll do that

templates/tensorflow-gpu-1.12/image_classifier/src/helpers/download_helpers.py
zeryx: I like doc strings but they aren't python 2 compatible normally (whcih we still support blah)

templates/tensorflow-gpu-1.12/image_classifier/src/helpers/download_helpers.py
zeryx: yea true not a bad idea, I don't think we've played around with "advanced" stuff in py such as *vars haha

templates/tensorflow-gpu-1.12/image_classifier/src/example_test.py
zeryx: I'll remove it for now, KISS.

j-vasanth: @zeryx wanted to put that test in as a best practice that we are promoting. Also our existing Hello app doeos have a test too, so it would be consistent. I am fine with either.

besirkurtulmus: Jai's point was my understanding too.

Unless we generate the standard `algo_test.py` somewhere else, I don't see why it shouldn't be included here either.

templates/tensorflow-gpu-1.12/src/ALGORITHM.py
```diff
@@ -0,0 +1,57 @@
...
+    else:
+        raise Exception("Input should be either a string or json object.")
+    local_image = get_image(client, image_url, image_dimensions)
+    prediction = predict(model, local_image, classes)
+    output = {'prediction': prediction}
+    return output
+
+
+if __name__ == "__main__":
+    input = {'image_url': "https://i.imgur.com/j2kloCx.jpg"}
```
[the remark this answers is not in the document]
zeryx: not a bad idea, I think using a real image is useful though - it helps someone who doesn't quite know how an algorithm flow should look, get a better intuitive understanding. Seeing a real image from a hosted location helps in learning. At least that's what I'd argue, no idea if true or not :man_shrugging:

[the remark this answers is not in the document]
zeryx: I will update this to be from a self hosted 3 bucket tho

templates/tensorflow-gpu-1.12/image_classifier/src/helpers/download_helpers.py
zeryx: I think this might be something we address in an update, closing for now but will make an issue for it in algorithmiaio/templates

templates/tensorflow-gpu-1.12/image_classifier/src/example_test.py
zeryx: up to you guys, I'm in different - what do you think @pmcq ?

pmcq: i felt that the comment was misleading since these tests aren't run during creation of a template or anything.  as far as the file goes, i don't really have a preference but should maybe follow our existing template https://github.com/algorithmiaio/langpacks/tree/master/python/template/src (i.e. called `__ALGO___test.py` (__ALGO__ will be substituted automatically)

templates/tensorflow-gpu-1.12/image_classifier/src/example.py
pmcq: i misspoke earlier it's just `__ALGO__.py` https://github.com/algorithmiaio/langpacks/tree/master/python/template/src

pmcq:
you'll also need an algorithmia.conf file (or the pipe script won't know what to load).  can just cp the file from https://github.com/algorithmiaio/langpacks/blob/master/python/template/algorithmia.conf to the same location in this template

zeryx:
haha yes, I use intellij/pycharm defaults, if other devs are using other IDE or no-IDE environments I def agree we should be consistent. @j-vasanth @besirkurtulmus what standards do you guys follow?

j-vasanth:
I am partly to blame for the inconsistency in the template repo, but I vote for double space indentation everywhere.

zeryx:
I vote for pycharm defaults, as it's the most popular IDE and most folks don't deviate. When I open one of your files that isn't tab based it throws a ton of soft warning messages everywhere.

j-vasanth:
PEP-8 recommends using spaces. https://www.python.org/dev/peps/pep-0008/#tabs-or-spaces

pmcq:
i think 4 spaces is usually the most common.  i also wasn't sure how the in-browser editor we have handles/renders tabs, but not trying to jump in the middle here haha

j-vasanth:
Yeah we agreed to using 4 spaces.

templates/tensorflow-gpu-1.12/src/helpers/download_helpers.py
```diff
@@ -0,0 +1,100 @@
+import pickle
+import urllib
+import os
+from Algorithmia.errors import AlgorithmException
+from tempfile import mkdtemp
```
j-vasanth: This looks like it is still tab indented. Change it to 4 spaces if it is.

templates/tensorflow-gpu-1.12/src/helpers/preprocessing_helpers.py
```diff
@@ -0,0 +1,23 @@
+from PIL import Image
+import numpy as np
```
j-vasanth: Change to 4 space indentation.

templates/tensorflow-gpu-1.12/src/ALGORITHM.py
```diff
@@ -52,6 +66,10 @@ def apply(input, client=CLIENT, model=MODEL, image_dimensions=TARGET_IMAGE_DIMEN
 
 
 if __name__ == "__main__":
-    input = {'image_url': "https://i.imgur.com/j2kloCx.jpg"}
+    """
+    Want to test your algorithm out locally for debugging purposes? Anything in this `if` code block is only
```
besirkurtulmus: There is no if code block below this doc string.

besirkurtulmus:
I've added the `install.sh` file for installing the template.

Let me know, want to get this merged in by today if possible.

templates/tensorflow-gpu-1.12/src/__ALGO___test.py
```diff
@@ -0,0 +1,15 @@
+from .__ALGO__ import apply
+
+"""
+This is the testing file, which should be bundled along with the Algorithm.py file when creating a template.
+These tests are run during the creation of a template, and if they fail - the templating process fails as well.
+"""
+
+def primariy_api_example():
+    input = {'image_url': 'https://s3.amazonaws.com/algorithmia-uploads/money_cat.jpg'}
```
neildf: would be nice if we could put this behind a DNS entry in the future. Not a blocker for this PR tho

libraries/tensorflow-gpu-1.12/install.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+set -e
+
+# Install the requirements defined in the requirements.txt file in the template
+pip install -r ../../templates/tensorflow-gpu-1.12/requirements.txt
+
+# Give algo user ability to write to updated keras/tensorflow files
+find $PYTHON_LIB_PATH/$PYTHON_VERSION/site-packages -user root | xargs chown algo:algo
```
neildf: This doesn't seem ideal, what's the use case?

templates/tensorflow-gpu-1.12/src/__ALGO___test.py
besirkurtulmus: I agree 100%

libraries/tensorflow-gpu-1.12/install.sh
besirkurtulmus: The pip installation above is done under root.

This would prevent a user to update their `tensorflow-gpu` version if they desire to do so.

pmcq: They shouldnt be using this image if they wanted a different version of tensorflow though, that would be a big waste of space. They could just use the normal pythin+cuda image. I'm also not certain those variables would be set.

libraries/tensorflow-gpu-1.12/install.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+set -e
+
+# Install the requirements defined in the requirements.txt file in the template
+pip install -r ../../templates/tensorflow-gpu-1.12/requirements.txt
```
pmcq: mentioned in chat but you can't refer to a file outside of this directory since it wont be in the docker context. You can put the file in a directory named "context" tho

libraries/tensorflow-gpu-1.12/install.sh
neildf: yeah, arguably that would defeat the purpose of the packageset. We should just have package sets per TF version

templates/tensorflow-gpu-1.12/src/__ALGO___test.py
j-vasanth: Created https://algorithmia.atlassian.net/browse/ALGO-531

libraries/tensorflow-gpu-1.12/install.sh
besirkurtulmus: This is with the assumption that we won't be able to change an algorithms packageset retroactively, otherwise the user would have to create a new algorithm from scratch.

But I guess that's an issue out-of-scope here.

I'll remove the additional command.

pmcq: Users will be able to change the package set (will take effect on next compile)

besirkurtulmus: Good to know.

libraries/tensorflow-gpu-1.12/install.sh
besirkurtulmus: Fixed this issue.

libraries/tensorflow-gpu-1.12/install.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+set -e
+
+# Install the requirements defined in the requirements.txt file in the template
+pip install -r context/requirements.txt
```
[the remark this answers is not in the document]
besirkurtulmus: We can change this to `pip3` if needed.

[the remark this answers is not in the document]
pmcq: yeah i think this will be okay to test and we'll know pretty fast whether it needs to be pip3 (but that would also mean that we would need a separate package to install tensorflow-1.12 for python2 too)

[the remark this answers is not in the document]
besirkurtulmus: Cool. We can change this easily.
We can add new a new package for python2 after testing for this is done.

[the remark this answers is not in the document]
besirkurtulmus: Also, if we're going to create a new package for python2, should we also include `py2` in the name of the packageset?

besirkurtulmus:
@pmcq sounds good. I'll merge it.

PR #39, opened 2018-12-20
Added files for python3 IPA image

adnaanm:
The pipe script could definitely be optimized since this is now specifically for python 3 but I didn't want to mess with it for the time being


languages/python3/template/requirements.txt
```diff
@@ -0,0 +1 @@
+algorithmia>=1.0.0,<2.0
```
pmcq: i think we still need `six` (it's in the pipe script) unless that's bundled with python 3.7 now?

languages/python3/template/algorithmia.conf
```diff
@@ -0,0 +1,4 @@
+{
+    "username": "__USER__",
+    "algoname": "__ALGO__"
```
pmcq: nit: would be good to have `language` in here still.  we haven't created it on the backend but probably `python3.7` would be our name

pmcq (approved):
looks good - my only thought is whether we should rename the libraries directory - will send out a poll in a minute on it

languages/python3/template/requirements.txt
adnaanm: I actually thought it was since it imports fine on my system and according to my terminal history I never installed it. The official standard library (https://docs.python.org/3/library/) doesn't seems to have it though so I'll add it back, thanks for the catch

languages/python3/template/algorithmia.conf
adnaanm: 👍

PR #40, opened 2018-12-21
ALGO-529 creating a static pytorch template

zeryx:
As mentioned here:
https://algorithmia.atlassian.net/browse/ALGO-529
We're trying to create some static package-set templates for a few ml frameworks, this is a proposed PR for the pytorch one.

### DIFF
--- a/templates/pytorch/requirements.txt
+++ b/templates/pytorch/requirements.txt
@@ -1,0 +1,5 @@
+algorithmia>=1.0.0,<2.0
+six
+numpy
+Pillow
+https://download.pytorch.org/whl/cu92/torch-1.0.0-cp35-cp35m-linux_x86_64.whl
--- a/templates/pytorch/src/__ALGO__.py
+++ b/templates/pytorch/src/__ALGO__.py
@@ -1,0 +1,57 @@
+import Algorithmia
+import torch
+from . import helpers
+from PIL import Image
+import numpy as np
+
+TARGET_IMAGE_SIZE = "224"
+CLASSES = "['cat', 'dog']"
+MODEL_PATH = "data://demo/pytorch_template/model.t7"
+
+client = Algorithmia.client()
+
+
+def get_model(client, model_path):
+    local_file = client.file(model_path).getFile().name
+    model = torch.jit.load(local_file)
+    return model
+
+
+def predict(model, image_path):
+    image = Image.open(image_path)
+    tensor = torch.Tensor(np.asarray(image))
+    tensor = tensor.view(1, 3, TARGET_IMAGE_SIZE, TARGET_IMAGE_SIZE)
+    result = model.forward(tensor)
+    arged_max = np.argmax(result[0])
+    pred_class = CLASSES[int(arged_max)]
+    return pred_class
+
+
+
+def get_image(client, image_url, image_dimensions):
+    local_image_path = helpers.download_helpers.download_image(client, image_url, size=image_dimensions)
+    return local_image_path
+
+
+def apply(input):
+    if isinstance(input, str):
+        image_url = input
+    elif isinstance(input, dict):
+        if 'image_url' in input:
+            image_url = helpers.download_helpers.type_check(input, 'image_url', str)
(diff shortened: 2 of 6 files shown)


templates/pytorch/requirements.txt
```diff
@@ -0,0 +1,5 @@
+algorithmia>=1.0.0,<2.0
+six
+numpy
+Pillow
```
besirkurtulmus: We might want to version peg this.

templates/pytorch/src/__ALGO__.py
```diff
@@ -0,0 +1,57 @@
+import Algorithmia
+import torch
+from . import helpers
+from PIL import Image
+import numpy as np
+
+TARGET_IMAGE_SIZE = 224
+CLASSES = ['cat', 'dog']
```
besirkurtulmus: nit: we're using single and double quotes in this file. I think we should only commit to one: single quotes?

templates/pytorch/src/__ALGO__.py
```diff
@@ -0,0 +1,57 @@
...
+   local_image = get_image(client, image_url, TARGET_IMAGE_SIZE)
+   prediction = predict(model, local_image)
+   output = {'prediction': prediction}
+   return output
+
+
+model = get_model(client, MODEL_PATH)
+
+if __name__ == "__main__":
+   input = {'image_url': "https://i.imgur.com/j2kloCx.jpg"}
```
besirkurtulmus: This can easily break. Image should be in an algorithmia s3 bucket instead.

templates/pytorch/src/helpers/download_helpers.py
```diff
@@ -0,0 +1,100 @@
+import pickle
+import urllib
+import os
+from Algorithmia.errors import AlgorithmException
+from tempfile import mkdtemp
+
+def download_image(client, image_url, size=None):
+	"""
```
besirkurtulmus: These are also in tabs, while the other files are in 4 space indentation.

I'll make a commit to fix this in this PR too.

templates/pytorch/requirements.txt
pmcq: fyi - if you don't version peg then if a version is already installed pip won't install a new one (which might be preferable since we will have _some_ version installed in the base pytorch image which won't ever change).  i'm kind of counting on you guys to know what pip packages need to be installed for those as well

besirkurtulmus: I'd rather have all these package be as explicit as possible.

We want to be able to reproduce things if needed.

@zeryx: I would just do a `pip freeze` to get the version that was installed on your machine when you tested this locally.

j-vasanth: But should we not be using the  package that was bundled into the package set (if one exists) rather then pulling something else ?  So if this python package already exists (and we know that it does) why specify it in requirement.txt at all?

pmcq: yeah that was my point as well - the package (which doesn't yet exist but needs to in `langpacks/libraries/pytorch-1.0`) would do some `pip install` in there and that is the base image from which any algorithm would be built.  if there are conflicting versions set in requirements.txt then the space saving we get from the base image would be lost

besirkurtulmus:
The following issues were addressed in the last few commits:

* Converted tabs into 4 spaces in appropriate places.
* Used the Python 3.7 version of PyTorch in the dependencies file.
* Renamed the `pytorch` directory to `pytorch1.0.0` to reflect version number.
* Version pegged the Pillow version in the dependencies file.
* Used the s3 link instead of a imgur link.
* Converted all of the double quotes into single quotes in appropriate places.
* Added missing `algorithmia.conf` file.
* Removed the main function from the algorithm script file.

besirkurtulmus:
I've added the `install.sh` file for installing the template.

Let me know if this looks good, want to get this merged in by today if possible.

besirkurtulmus:
@pmcq sounds good here too. I'll merge it.

PR #41, opened 2018-12-22
Added more information based on my understanding

adnaanm:
I added some info based on my current understanding that I think would give Denzel and others new to what langpacks does some more context on how it works and how to add new languages. Please let me know if any additions or changes are incorrect or misleading, I tried to be as precise as possible.


pmcq (approved):
:clap: :clap: :tada: :beer:  thanks!

PR #42, opened 2019-01-01
A number of fixes to the python3 setup

pmcq:
Fixing the templates and install scripts (by moving to the install scripts from the official python3.7 docker images).  it's mit licensed so I believe including the copyright is all i need to do here.


libraries/python37/Dockerfile
```diff
@@ -0,0 +1,94 @@
...
+
+# make some useful symlinks that are expected to exist
+RUN cd /usr/local/bin \
+    && ln -s idle3 idle \
+    && ln -s pydoc3 pydoc \
+    && ln -s python3 python \
+    && ln -s python3-config python-config
+
+# if this is called "PIP_VERSION", pip explodes with "ValueError: invalid truth value '<VERSION>'"
+ENV PYTHON_PIP_VERSION 18.1
```
neildf: any idea why this is?

PR #43, opened 2019-01-04
Use a different keyserver to hopefully prevent sporadic errors seen in our build environment

pmcq:
In testing there seems to be some sporadic errors when trying to get the GPG key when installing python, trying to use a different keyserver address here based on some google-fu

### DIFF
--- a/libraries/python37/Dockerfile
+++ b/libraries/python37/Dockerfile
@@ -34,7 +34,7 @@ RUN set -ex \
     && wget -O python.tar.xz "https://www.python.org/ftp/python/${PYTHON_VERSION%%[a-z]*}/Python-$PYTHON_VERSION.tar.xz" \
     && wget -O python.tar.xz.asc "https://www.python.org/ftp/python/${PYTHON_VERSION%%[a-z]*}/Python-$PYTHON_VERSION.tar.xz.asc" \
     && export GNUPGHOME="$(mktemp -d)" \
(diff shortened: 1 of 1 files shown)


libraries/python37/Dockerfile
```diff
@@ -28,12 +28,13 @@ ENV PATH /usr/local/bin:$PATH
...
 ENV PYTHON_VERSION 3.7.1
 
+# Setting keyserver per https://github.com/tianon/gosu/issues/17#issuecomment-203074866
 RUN set -ex \
     \
     && wget -O python.tar.xz "https://www.python.org/ftp/python/${PYTHON_VERSION%%[a-z]*}/Python-$PYTHON_VERSION.tar.xz" \
     && wget -O python.tar.xz.asc "https://www.python.org/ftp/python/${PYTHON_VERSION%%[a-z]*}/Python-$PYTHON_VERSION.tar.xz.asc" \
     && export GNUPGHOME="$(mktemp -d)" \
-    && gpg --batch --keyserver ha.pool.sks-keyservers.net --recv-keys "$GPG_KEY" \
+    && gpg --batch --keyserver hkp://p80.pool.sks-keyservers.net:80 --recv-keys "$GPG_KEY" \
```
jamesatha: If there is a https option, why not?
hkps://hkps.pool.sks-keyservers.net

pmcq:
womp womp... getting  `gpgkeys: HTTP fetch error 1: unsupported protocol` with `hkps://hkps.pool.sks-keyservers.net` i'm going to switch back or try experimenting to see what works correctly - it's a bit roundabout to test this

PR #45, opened 2019-01-06
Fixups to the compile dockerfile 

pmcq:
want to make sure the group is also set to algo for all files, and since the builder/runner images are already end with USER algo, trying to change the user in the dockerfile results in permissions errors

### DIFF
--- a/languages/Dockerfile.compile.j2
+++ b/languages/Dockerfile.compile.j2
@@ -1,12 +1,10 @@
 FROM {{builder_image}} as builder
-USER algo
-COPY --chown=algo algosource /opt/algorithm/
+COPY --chown=algo:algo algosource /opt/algorithm/
 RUN /opt/algorithm/bin/build
 
 FROM {{runner_image}}
 {% for artifact in config.artifacts %}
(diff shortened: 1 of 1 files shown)


jamesatha:
This PR seems fine. But I'm also a bit confused about how this helps. Would it be any better to take out the change to USER algo from the builder and runner. We only really care about setting the user to algo in the compile docker file before we run /opt/algorithm/bin/build. And it should be last thing we do after we copy all the artifacts.

pmcq:
I think we would still want to be user algo from the beginning.  It would be a pretty big security problem to run the /bin/build script as root, and in case we allow pulling the builder/runner images locally or want to have other ways to build the algorithm images it'd be one more thing to remember. :shrug:

pmcq:
Turns out there was an issue with the way kaniko was building things - the rootfs from the base image of the runner wasn't being unpacked since it didn't think it had to unpack that layer, and hence /etc/passwd literally couldn't be found.  I fixed our fork of kaniko and it's working now with no changes to this template - https://github.com/pmcq/kaniko/commit/04ea1cd20b5407f626e4ff485b7684f138c3c2c3

PR #47, opened 2019-01-08
Use numeric uids/gids instead of named user so that user id lookups aren't required and therefore can use the kaniko cache effectively

pmcq:
The change requires a bump in the version of kaniko in use (https://github.com/algorithmiaio/kaniko/pull/2) but I also have a couple questions to resolve via committee here:
1. The builder/runner images already end with the `USER algo` lines so having it in the compile dockerfile is a bit unnecessary
2. Choice of 2222 was totally arbitrary but didn't know if others were aware of this - https://github.com/algorithmiaio/langpacks/blob/master/languages/Dockerfile.builder.j2#L10.  I chose it because it seemed like a good idea to not be the same as users that might already exist on our system, and because we've had confusing situations in the past with local development where something works for one person and not another person because the UID would be 1000 or something, but some people's desktop UIDs are 1001 etc.
3. We could also have the algo_uid be a variable that legit adds in to builder/runner/compile images but that felt a bit like overkill.  All that matters is that the ID is the same in the 3 images (which they are fixed here).

### DIFF
--- a/languages/Dockerfile.compile.j2
+++ b/languages/Dockerfile.compile.j2
@@ -1,6 +1,7 @@
 FROM {{builder_image}} as builder
 USER 2222
 COPY --chown=2222 algosource /opt/algorithm/
+ENV HOME=/home/algo
 RUN /opt/algorithm/bin/build
 
 FROM {{runner_image}}
--- a/languages/Dockerfile.compile.j2
+++ b/languages/Dockerfile.compile.j2
@@ -1,13 +1,13 @@
 FROM {{builder_image}} as builder
-USER 2222
-COPY --chown=2222 algosource /opt/algorithm/
+USER $ALGO_UID
+COPY --chown=$ALGO_UID:$ALGO_UID algosource /opt/algorithm/
 ENV HOME=/home/algo
 RUN /opt/algorithm/bin/build
 
 FROM {{runner_image}}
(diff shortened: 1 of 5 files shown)


pmcq:
there's no longer any bindmounting needed in the construction of the base images or algo images :) .  we do have to bindmount the cache into the kaniko container but that's totally separate from this here

languages/Dockerfile.compile.j2
```diff
@@ -1,12 +1,13 @@
 FROM {{builder_image}} as builder
-USER algo
-COPY --chown=algo algosource /opt/algorithm/
+USER 2222
+COPY --chown=2222 algosource /opt/algorithm/
+ENV HOME=/home/algo
```
jamesatha: little curious why env var needed to be set

pmcq: pip install --local was otherwise trying to write to /.local and it doesnt have permissions to do so.

pmcq: Kaniko sets the value of HOME to / if the user is a UID and not a username - https://github.com/algorithmiaio/kaniko/blob/master/pkg/commands/run.go#L134-L141.  I'm actually still poking around with this because there's some weird file permissions on /home/algo in some cases (depending on whether caching was used and whether the image was in the cache or not) :(

PR #49, opened 2019-01-11
Copy Python37 install to Python36 directory

pmcq:
Also adding a small tweak to the python runtime so pip-installed binaries are available on PATH.  I will be testing this in test and might have to submit tweaks but it's a little easier to test this using test infra if I merge first

### DIFF
--- a/libraries/python36/Dockerfile
+++ b/libraries/python36/Dockerfile
@@ -1,0 +1,95 @@
+# Lightly modified from https://github.com/docker-library/python/blob/ab8b829cfefdb460ebc17e570332f0479039e918/3.7/stretch/Dockerfile
+
+# MIT license from initial repo
+# Copyright (c) 2014 Docker, Inc.
+#
+# Permission is hereby granted, free of charge, to any person obtaining
+# a copy of this software and associated documentation files (the
+# "Software"), to deal in the Software without restriction, including
+# without limitation the rights to use, copy, modify, merge, publish,
+# distribute, sublicense, and/or sell copies of the Software, and to
+# permit persons to whom the Software is furnished to do so, subject to
+# the following conditions:
+#
+# The above copyright notice and this permission notice shall be included
+# in all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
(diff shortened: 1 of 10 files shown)


libraries/python36/Dockerfile
```diff
@@ -0,0 +1,95 @@
...
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+# EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+# MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+# IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+# CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+# TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+# SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+# ensure local python is preferred over distribution python
+ENV PATH /usr/local/bin:$PATH
```
jamesatha: random comment. this is interesting/could cause confusion later if a later package also needs to update the PATH which puts the original python path

libraries/python36/Dockerfile
```diff
@@ -0,0 +1,95 @@
...
+
+RUN set -ex; \
+    \
+    wget -O get-pip.py 'https://bootstrap.pypa.io/get-pip.py'; \
+    \
+    python get-pip.py \
+        --disable-pip-version-check \
+        --no-cache-dir \
+        "pip==$PYTHON_PIP_VERSION" \
+    ; \
```
jamesatha: we have `-e` in both RUN commands so the `&&`s seem unnecessary. But can we be consistent and use `&&` or `;` everywhere?

Also, I comment above saying what we are `rm`-ing would be nice.

(Neither of these are a huge deal since this is copy/pasted from the other one)

pmcq: yeah i'd rather leave it is just copy/pasted code than change the style like that

libraries/python36/Dockerfile
pmcq: not sure I understand your comment (but this was also part of the copy/paste).  there isn't actually a version of python installed in the base ubuntu:16.04 images or anything

PR #54, opened 2019-01-18
 PLATFORM-2484 IPA Testing utilities

zeryx:
working with @adnaanm to build this out to simplify the testing/validation process.

### DIFF
--- a/tools/ipa_packageset_builder.py
+++ b/tools/ipa_packageset_builder.py
@@ -37,7 +37,7 @@ import argparse
 if __name__ == "__main__":
     parser = argparse.ArgumentParser(description='Creates a packageset dockerfile, by combining package templates together.')
     parser.add_argument('--base-image', dest='base_image', type=str, required=True)
-    parser.add_argument('--packages', dest='packages', type=list, required=True)
+    parser.add_argument('-p', '--package', action='append',  dest='packages', required=True)
     args = parser.parse_args()
     print(args.base_image)
     print(args.packages)
--- a/tools/ipa_packageset_builder.py
+++ b/tools/ipa_packageset_builder.py
@@ -11,17 +11,12 @@ LANGSERVER_IMAGE ="algorithmiahq/langserver:{}".format(LANGSERVER_VERSION)
 RUNNER_PATH = path.join(DIR_PATH_TO_TEMPATES, RUNNER_NAME)
 
 class Package:
-    def __init__(self, package_name, install_script, dockerfile_path, base_image):
+    def __init__(self, package_name, install_script, dockerfile_path):
         self.script = install_script
         if dockerfile_path:
             self.dockerfile = get_dockerfile_as_string(dockerfile_path)
         else:
             raise Exception('dockerfile path not available for package {}'.format(package_name))
-        if base_image:
-            self.base_image = base_image
-        else:
-            raise Exception('base_image path not available for package {}'.format(package_name))
-
 
 
 def get_dockerfile_as_string(file_path):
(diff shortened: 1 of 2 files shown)


tools/ipa_packageset_builder.py
```diff
@@ -0,0 +1,82 @@
...
+        packages=packages,
+        base_image=base_image,
+        langpacks_version='',
+        langserver_image=LANGSERVER_IMAGE)
+    save_generated_template(generated_template, output_file_path)
+
+    print("completed template construction, file available at {}".format(output_file_path))
+
+
+"""Idk if you wanna use argparse but it's p easy to use."""
```
pmcq: yup would always want to use argparse

tools/ipa_packageset_builder.py
```diff
@@ -0,0 +1,82 @@
...
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description='Creates a packageset dockerfile, by combining package templates together.\n'
+                                                 'Make sure to run this from the root directory.')
+    parser.add_argument('-b', '--base-image', dest='base_image', type=str, required=True)
+    parser.add_argument('-o', '--output-filename', dest='output_path', required=True)
+    parser.add_argument('-p', '--package', action='append',  dest='packages', required=True)
+    args = parser.parse_args()
+    print(args.base_image)
```
pmcq: nit: are these prints really necessary

tools/ipa_packageset_builder.py
```diff
@@ -0,0 +1,82 @@
+from jinja2 import Template
```
pmcq: can you add a `#!/usr/bin/env python3` to the top here

tools/ipa_packageset_builder.py
```diff
@@ -0,0 +1,82 @@
...
+        return filepath
+    else:
+        return None
+
+def build(base_image, package_dirs, output_file_path):
+
+    raw_template = get_template(RUNNER_PATH)
+    packages = []
+    for dir in package_dirs:
+        dockerfile_path = path.join('',  *[DIR_PATH_TO_PACKAGES, dir, "Dockerfile"])
```
pmcq: nit: can't you just do `path.join('', DIR_PATH_TO_PACKAGSE, dir, "Dockerfile")`?

tools/ipa_packageset_builder.py
```diff
@@ -0,0 +1,82 @@
...
+RUNNER_PATH = path.join(DIR_PATH_TO_TEMPATES, RUNNER_NAME)
+
+class Package:
+    def __init__(self, package_name, install_script, dockerfile_path, base_image):
+        self.script = install_script
+        if dockerfile_path:
+            self.dockerfile = get_dockerfile_as_string(dockerfile_path)
+        else:
+            raise Exception('dockerfile path not available for package {}'.format(package_name))
+        if base_image:
```
pmcq: i don't think packages need the `base_image`?

tools/ipa_packageset_builder.py
zeryx: Nah I kept them for diagnostics, removing em

tools/ipa_packageset_builder.py
zeryx: Does that work? I didn't think you could do that with path.join, I'll give er a shot

tools/ipa_packageset_builder.py
zeryx: the template doesn't have logic to remove the `FROM {{base_image}}` like an if check, so I assumed it was mandatory.

tools/ipa_packageset_builder.py
zeryx: sweet, will continue to use then

tools/ipa_packageset_builder.py
pmcq: it needs to be in the context but it's not a property on each package that is referenced (it seems to properly be in the context down below)

zeryx:
Can I merge this without too much pain? @pmcq @neildf

pmcq (approved):

PR #56, opened 2019-01-21
added mxnet/gpu/cuda9.2 package to libraries

zeryx:
this is MVP for getting mxnet gpu support functional on our platform.

is compatible with https://github.com/algorithmiaio/langpacks/pull/54

### Requirements for building with the packagset_builder.py script:
* base_image - nvidia/cuda:9.2-devel
* packages - python36 or python37
### Requires for local testing:
* nvidia docker
   * you can enter into and test with GPU support after docker image is built with `docker run --rm --runtime="nvidia" --entrypoint /bin/bash mxnet_packageset_name:latest .`
   * Assuming your system has a dedicated nvidia GPU of course..


pmcq:
Our hosts only support up to cuda 9.1, and other images i've generated have been off of nvidia/cuda:9.0 is it possible to use cuda 9.0?

zeryx:
yea we can use 9.0, I just don't have 9.0 on my local system, I can adjust.

pmcq:
the cuda versions are supposed to be backward compatible so if you have cuda9.2 installed you can also run 9.0 (the nvidia/cuda:9.0 base image has the runtime libraries for 9.0 included)

pmcq:
Yes the host does need CUDA installed and the nvidia driver.  The nvidia driver only supports up to a certain version of CUDA (9.1 in our case) and the CUDA driver only supports up to 9.1 as well, but is backward compatible with only versions of the runtime packages (which are what is installed in the nvidia/cuda:9.x images).  If the docker image requests a CUDA version that is not supported by the host then it will fail to start - see https://devblogs.nvidia.com/gpu-containers-runtime/

zeryx:
Do we have a system in place to roll out driver updates to our machines? Or is that still going to remain static? I can see a world where that can cause us similar problems down the road.

templates/mxnet-cu90-1.3.1/requirements.txt
```diff
@@ -0,0 +1,4 @@
+algorithmia>=1.0.0,<2.0
+six
```
pmcq: minor: as of yesterday you don't have to include `six` in here anymore since it will be installed with the pipe script (which is what actually needs it in the first place).

pmcq (approved):
seems good to me - do you know if it mxnet in pypi supports python3.6 and 3.7?

PR #57, opened 2019-01-24
simplifying the tensorflow template

zeryx:
the current tensorflow template is great but it is a bit too complex, it requires both internet access _and_ recursive algo call permissions.

This new version is a simple matrix multiplication algorithm, that tests the gpu and makes sure everything works properly.


templates/tensorflow-gpu-1.12/requirements.txt
```diff
@@ -1,5 +1,3 @@
 algorithmia>=1.0.0,<2.0
 six
-keras==2.1.4
-tensorflow-gpu==1.12.0
-pytest==4.0.2
\ No newline at end of file
+tensorflow-gpu==1.12.0 # if you need a different version, please change this value - bare in mind that it may take longer to compile.
```
kennydaniel: should be "bear in mind" :bear:

zeryx: :bear: has been added

kennydaniel:
In the interest of making it SLIGHTLY more illustrative, how would you feel about actually using the input? Like maybe if the input is a matrix, multiply it by itself (or its transform)?

eg (pseudocode, but do it with keras?):
```
A = np.array(input)
return A.dot(A.transform())
```

zeryx:
Was thinking about that, having it do nothing keeps it familiar with normal "hello world" stuff - but I'm certainly open to actually using the input.

zeryx:
Requiring the user to provide a matrix input is a bit cumbersome, that adds a lot of complexity from a first time user UI perspective I feel. I was thinking that the user input value could adjust the shape of the random vectors that are multiplied together, and take the dot of them

zeryx:
I do agree that the dot of the inverse might be nice..

kennydaniel:
Yea could also just pass in dimensions. Just trying to think what will be the most instructive to users. another proposal:
```
{
  "matrix1": [[1,0],[0,1]],
  "matrix2": [[2,3],[4,5]]
}
```
output:
```
{
  "dot": [[2,3], [4,5]]
}
```

zeryx:
I like that, has a bit of complexity and requires some input checking

zeryx:
> Looks good. General question: what happens if dimensions don't match? Is error sane?

yea that's why I added that actually, first time I tried with (20) and (30) and it failed with mismatched tensors. Might have been something I did though? Will double check before merging.

PR #58, opened 2019-01-24
Platform 2297 ipa bin build security

adnaanm:
Bin/build security fix for IPA algos; build script is now placed in and used from `/usr/local/bin` rather than the working directory

Work to be done
- Legit filter function


jamesatha:
I feel like it would be nicer if we renamed the files in context to reflect their new names but not 100% sure about that. It was nice that file names didn't change

PR #60, opened 2019-01-24
creating a pytorch template for a GPU enabled packageset

zeryx:


templates/pytorch-1.0.0/src/__ALGO__.py
```diff
@@ -1,53 +1,15 @@
...
-   return output
+    A = th.randn(input).cuda()
+    B = th.randn(input).cuda()
+    C = th.dot(A, B)
+    random_number = C.cpu().numpy().tolist()
+    return "with a random vector shape of {}, here is your random number {}".format(input, str(random_number))
 
 
-model = get_model(client, MODEL_PATH)
+print(apply(56))
```
pmcq: nit: probably don't want this

pmcq (commented):
seems fine to me too - maybe want a comment saying the expected input type is an int or something

pmcq (approved):

kennydaniel:
Holding off on approving this... probably want same behavior as tf example with matrix dot?

templates/pytorch-1.0.0/src/__ALGO__.py
zeryx: oh yea good point, will also remove that global debugging print

PR #61, opened 2019-01-24
Pulling out the bin/pipe to /usr/local/bin/

jamesatha:
Branched from Adnaans PR so this should look different when that goes in


adnaanm:
Since mine's been approved, I could close my PR and then you'd just merge this? This has all of my changes

PR #63, opened 2019-01-28
ALGO-532 - supporting an AllenNLP package.

zeryx:
This package depends on another framework package: `spacy-2.0.18`. 
Also requires a `python` package to build.


zeryx:
Hey @besirkurtulmus I think this is fully ready for an allen nlp tutorial template! I'm not sure how to create one honestly, there doesn't seem to be any simple tutorials or code snippets around.
If you can help with this that would be awesome!

kennydaniel:
Seems this is not yet ready for review?

zeryx:
No, should I remove reviewers until it's ready? I'm not 100% sure on etiquette here

templates/allennlp-0.8.1/requirements.txt
```diff
@@ -0,0 +1,6 @@
+algorithmia>=1.0.0,<2.0
+six
+numpy
+spacy==2.0.18 # if you change this, the algorithm will take longer to run and compile
```
kennydaniel: Do we want to specify versions here or not? I think we should decide on a policy for this. My understanding is that it won't reinstall if the version baked in the images is the EXACT same as the version listed here... but if we ever want to update the underlying IPA packagesets, it could leave users double-installing? @pmcq thoughts?

pmcq: so my thought is that the package is called `allennlp-0.8.1` for a reason (which is that the versions are fixed) and similarly this template name means it's for packageset `allennlp-0.8.1`.  my intuition was that we would create new packagesets for newer versions since things are usually incompatible, but if we wanted to bump revisions or anything then the package-set should be called `allennlp-0.8` and the algorithm templates should NOT have the versions specified so that the packageset could be modified.

pmcq: but i definitely am for the libraries (packages) to have explicit versions since otherwise builds would be indeterminate.  We do tag built images with githashes from langpacks so it would be okay to have a library `allennlp-0.8` and then change the install script from 0.8.1 -> 0.8.2 at some point (i.e. we would have to do a rebuild but any image we have built would be traceable back to the github so we could see when it was changed/that the different version is in place - tho checking package version when you import it would tell you too)

pmcq: but i don't know anything about this specific package so don't know what would make the most sense here.  are revision versions safe to update? are minor versions stable? etc.  i think it's something that will be package-dependent with some general advice

besirkurtulmus: @pmcq to answer your question about minor versions being stable, the answer is no. I think they actually broke one of their models in the version.

I'm more in favor of version tagging things, and creating new packagesets for every new version.

PR #64, opened 2019-01-28
ALGO-567 spacy package creation

zeryx:
Should be fully functional as-is, needed as a pre-requisite for allenNLP and other NLP based projects (such as some of EY's stuff)

### DIFF
--- a/libraries/spacy-2.0.18/Dockerfile
+++ b/libraries/spacy-2.0.18/Dockerfile
@@ -1,0 +1,9 @@
+
+RUN pip install spacy==2.0.18
+RUN python -m spacy download en
+RUN python -m spacy download de
+RUN python -m spacy download es
+RUN python -m spacy download pt
+RUN python -m spacy download fr
+RUN python -m spacy download it
+RUN python -m spacy download nl
--- a/templates/spacy-2.0.18/algorithmia.conf
+++ b/templates/spacy-2.0.18/algorithmia.conf
@@ -1,0 +1,5 @@
+{
+    "username": "__USER__",
+    "algoname": "__ALGO__",
+    "language": "python-3.x-1"
+}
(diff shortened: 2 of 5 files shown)


zeryx:
should  be ready for review

besirkurtulmus:
@zeryx any reason not to include every language that's available on spacy?

zeryx:
That should be all of the languages, did I miss one? the `xx` lang doesn't actually work very well from what I could see.

zeryx:
I intentionally only added the small models instead of the large ones, we can split up the package into small/large if we want to.

besirkurtulmus:
@zeryx, I see. My comment was based on your docstring: `You may change the language at runtime, but bear in mind that you'll get hit with some performance loss.`

Didn't realize that you added all available languages.

besirkurtulmus:
@zeryx, what's the size difference between using small & large models?

zeryx:
https://spacy.io/models/en
35 MB / 115 MB / 812 MB

They don't talk much about using the advanced models, and I feel like for most use cases the small's will be sufficient - but I could be persuaded to create a second spacy package that installs the large models instead.

besirkurtulmus:
@zeryx, I see. Definitely makes sense to add them later.

One question I have though is how do we want to add them? All in a single package, or individually?

besirkurtulmus (approved):
LGTM.

zeryx:
> @zeryx, I see. Definitely makes sense to add them later.
> 
> One question I have though is how do we want to add them? All in a single package, or individually?
Good question, I don't think we gain much from breaking them up, at least now. But for large models, docker image size is a concern. I'll have to defer to @pmcq and @neildf for that one.

pmcq:
i don't really have a preference without knowing things more specifically, but seems like all the small models are like 30MB which isn't much.  if the image layers are like, less than 400MB i wouldn't super worry about it, but we'll have to try it out and see too. i don't know anything about this tool or what people might typically end up using from it

zeryx:
I think most folks are only going to need the small models - DL doesn't make much sense for NLP.

kennydaniel (approved):

PR #65, opened 2019-01-31
enabling pytorch full functionality w/o reqs.txt

zeryx:
As @dinoboy197 noticed, the pytorch image isn't feature complete, as it needs numpy to work properly - pytorch uses numpy but doesn't install it automatically for some reason.

This isn't an issue from a users perspective, but it is a problem from a testing perspective, and it makes my "how to make & test an IPA package" tutorial a bit flimsy.

By adding numpy here, it will mean that an IPA dev will be able to fully test a pytorch project inside of a docker environment without needing to `pip install -r requirements.txt` like our platform does presently.

Alternatively, we could make a `docker run` script that automatically executes `pip install -r requirements.txt` and runs the `__ALGO___test.py` script automatically. That could be a nice improvement but it would require some development time.

### DIFF
--- a/libraries/pytorch-1.0.0/Dockerfile
+++ b/libraries/pytorch-1.0.0/Dockerfile
@@ -1,1 +1,2 @@
+RUN pip install numpy==1.16.0
 RUN pip install torch==1.0.0


pmcq (approved):
makes sense to me - i would have expected torch to install the numpy properly too but if it doesn't then this would be the right behavior

besirkurtulmus (approved):
I do think the right way to test things would be to install the `requirements.txt` file, like James said in his last paragraph.

But if we're looking for a quick solution, this LGTM.

PR #66, opened 2019-01-31
ALGO-569  C# .net 2.2 IPA language support

zeryx:
This task encompasses getting .net 2.2 language support fully operational with our new IPA package build system.


libraries/csharp-net22-runtime/context/Pipe/Module.cs
```diff
@@ -0,0 +1,94 @@
...
+                    {
+                        throw new Exception($"Invalid Json, expected field '{field.Name}' to be defined.");
+                    }
+                }
+
+                return ApplyMethod.Invoke(null, new[] {deserializedObj});
+            }
+            else
+            {
+                throw new Exception($"content_type: '{request.ContentType}' is not implemented!");
```
zeryx: As per Kenny's advice, for this MVP I've only implemented `json` as a request/response type.

libraries/csharp-net22-runtime/context/Pipe/Program.cs
```diff
@@ -0,0 +1,53 @@
...
+            Module algoModule;
+            try
+            {
+                algoModule = new Module(config);
+            }
+            catch (Exception e)
+            {
+                ExceptionResponse response = new ExceptionResponse(e);
+                Write.WriteJsonToPipe(response);
+                Console.Out.WriteLine("PIPE_INIT_FAILED");
```
zeryx: I feel like this is a good thing to add? I can remove if it would compilcate things ofc.

libraries/csharp-net22-runtime/context/Pipe/Module.cs
kennydaniel: may want to handle type "text" too.

libraries/csharp-net22-runtime/context/Pipe/Program.cs
kennydaniel: I think that will break json parsing of the output?

zeryx:
I still need to add a `bin/build` and `bin/test`, those are the two components I'll add today.
I'd also very much like to add unit tests to some critical functions however I'm not quite sure how to do that in C# yet... will look into it.

libraries/csharp-net22-runtime/context/Pipe/Module.cs
zeryx: yea `text` is pretty easy to support, I was thinking the same.

libraries/csharp-net22-runtime/context/Pipe/Program.cs
zeryx: I thought stdout was used for logging, where `/tmp/algoout` was the primary output response destination, no? I'm kind of flying blind with this so the smart thing to do would be to remove this.

zeryx:
Ok, current form seems to make sense however I don't know where the linkage is with langserver. Could use help with modifying IPA langserver to support non-python modules. It's certainly possible that I missed something. I'm not quite sure how to fully test, or if the scripts make sense based on pwd, etc.

libraries/csharp-net22-runtime/context/Pipe/Module.cs
```diff
@@ -0,0 +1,95 @@
...
+        }
+        private void LoadClass(Config config)
+        {
+            Assembly asm = Assembly.LoadFrom(DllPath);
+            PrimaryClassType = asm.GetType($"{config.Algoname}.{config.Algoname}");
+        }
+        private void GetApplyMethod()
+        {
+            // getting only public methods, as those are apply methods
+            MethodInfo applyMethod = PrimaryClassType.GetMethod("apply");
```
jamesatha: In Java we allow more than 1 and try out multiple. Why are we limiting ourselves to 1? Reflection is hard and maybe the wrong choice here and would like to discuss that more to make sure we build something robust

libraries/csharp-net22-runtime/context/Pipe/Module.cs
pmcq: IMO the MVP should conform to the spec that we have which includes `binary` I don't expect it would add much to support binary since it's just a b64 decode operation.  I don't know when else we are going to work on this if we don't do it correctly now (and otherwise will have deployed something broken which will need fixing anyway).  but yeah for initial local testing starting with json or whatevs is fine

libraries/csharp-net22-runtime/context/Pipe/Program.cs
pmcq: it doesn't add anything, and nothing will be trying to parse this so i think it should just be left out.

libraries/csharp-net22-runtime/context/Pipe/Module.cs
pmcq: i think the java reflection stuff is actually super brittle and doesn't do the write thing in a couple of cases (it certainly used to be the case, i'm not sure if it still is).  i think part of the problem with java was that like booleans could get coerced into strings, and also that to pass multiple arguments into a java method we had to like call it with an array like `[1, "foo", false]` which feels a little less easy to understand than the json blob kind of inputs we usually give for python stuff `{"count":1, "name": "foo", "debug": false}`.  so yeah i think it's worth a discussion to figure out what approach would be the best.  i forget what rust does off the top of my head - @anowell ?

libraries/csharp-net22-runtime/context/setup
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
```
pmcq: we don't need a setup script for new languages - don't see this in use anywhere

libraries/csharp-net22-buildtime/context/build
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+set -e
+
+dotnet restore
+dotnet build
```
pmcq: do you know if there's a way to set a mirror for dotnet packages or if that's just something users set in their project file?

libraries/csharp-net22-runtime/context/Pipe/Program.cs
```diff
@@ -0,0 +1,53 @@
...
+            while ((readLine = Console.In.ReadLine()) != null && readLine != "")
+            {
+                Request request = new Request(readLine);
+                object response = null;
+                try
+                {
+                    object result = algoModule.AttemptExecute(request);
+                     response = new Response(result, "json");
+                }
+                catch (Exception e)
```
pmcq: i'm not sure if there's a "non-fatal" exception we should be catching here (like how catching throwable in java is bad)

libraries/csharp-net22-runtime/context/Pipe/Program.cs
```diff
@@ -0,0 +1,53 @@
...
+                {
+                    response = new ExceptionResponse(e);
+                }
+                finally
+                {
+                    Write.WriteJsonToPipe(response);
```
pmcq: should validate that response is not null before writing it to the pipe.  if it is null then fill in a default error message

libraries/csharp-net22-runtime/context/.gitignore
```diff
@@ -0,0 +1 @@
+*.DotSettings.user
```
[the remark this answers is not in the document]
zeryx: honestly that's just my  IDE being weird, we can just remove that

libraries/csharp-net22-runtime/context/Pipe/.gitignore
```diff
@@ -0,0 +1,33 @@
+# Common IntelliJ Platform excludes
+
+# User specific
+**/.idea/**/workspace.xml
```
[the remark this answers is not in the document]
zeryx: no, again just my IDE

libraries/csharp-net22-runtime/context/Pipe/Program.cs
zeryx: ok I'll remove.

languages/csharp-dotnet-2.2/template/__ALGO__.csproj
```diff
@@ -0,0 +1,12 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp2.2</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Algorithmia.Client" Version="1.0.5" />
```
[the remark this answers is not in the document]
zeryx: I think we can wildcard but I'll double check.

libraries/csharp-net22-runtime/context/Pipe/Program.cs
```diff
@@ -0,0 +1,53 @@
...
+            
+            string readLine;
+            while ((readLine = Console.In.ReadLine()) != null && readLine != "")
+            {
+                Request request = new Request(readLine);
+                object response = null;
+                try
+                {
+                    object result = algoModule.AttemptExecute(request);
+                     response = new Response(result, "json");
```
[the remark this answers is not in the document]
zeryx: rider has some really weird indentation rules for C#.. I'll see if I can throw the whole thing into a linter

libraries/csharp-net22-runtime/context/Pipe/Module.cs
```diff
@@ -0,0 +1,95 @@
...
+            GetAlgoInputType();
+        }
+        
+        
+        //TODO: Figure out if this is always the way to capture the dll file of a project. Are there other project formats?
+        private void GetDllPath(Config config)
+        {
+            string algoName = config.Algoname;
+            string algoPath = config.Algopath;
+            string boilerplate = "bin/Debug/netcoreapp2.2";
```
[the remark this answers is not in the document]
zeryx: because I don't know how to create a release package yet. This is kind of designed to be a PoC that we can evaluate on test, and then build the real final version.

libraries/csharp-net22-runtime/context/Pipe/Module.cs
```diff
@@ -0,0 +1,95 @@
...
+            string fullpath = $"{algoPath}/{boilerplate}/{algoName}.dll";
+            Console.WriteLine(fullpath);
```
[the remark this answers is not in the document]
zeryx: Yea a macro is a good idea

libraries/csharp-net22-runtime/context/Pipe/Module.cs
```diff
@@ -0,0 +1,95 @@
...
+
+        private void GetAlgoInputType()
+        {
+            ParameterInfo[] parameters = ApplyMethod.GetParameters();
+            if (parameters.Length == 0)
+            {
+                throw new Exception("the discovered Apply method for your Algorithm did not have any input arguments!");
+            }
+
+            ParameterInfo paramInfo = parameters[0];
```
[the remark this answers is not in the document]
zeryx: Actually my thought here is that we're assuming the other parameters are optional, with defaults! Brian had an idea of having a `public static Output apply(SomeType input, Client client = Algorithmia.client())` apply method. That wouldn't effect anything.

Also, in the hopes of keeping this simple for testing, this is by far the easiest way to get things working. Assume an airity of 1, a single apply.. all json in/out.. etc..

libraries/csharp-net22-buildtime/context/build
zeryx: you can define it in the csproj file, yea you can also use private repos and whatnot. the main package mgmt system is managed here by the `__ALGO__.csproj` file.

languages/csharp-dotnet-2.2/config.json
```diff
@@ -0,0 +1,5 @@
+{
+    "artifacts": [
+        {"source":"/opt/algorithm", "destination":"/opt/algorithm/"}
```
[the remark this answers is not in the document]
zeryx: I think you only need the DLL file but I'll need to verify. This should work as is, but might be a bit heavy compared to what we need.

libraries/csharp-net22-runtime/context/Pipe/Program.cs
```diff
@@ -0,0 +1,53 @@
...
+                catch (Exception e)
+                {
+                    response = new ExceptionResponse(e);
+                }
+                finally
+                {
+                    Write.WriteJsonToPipe(response);
+                }
+            }
+            Console.WriteLine("PIPE_TERMINATE");
```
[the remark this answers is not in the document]
zeryx: I feel like it should be? I can remove them / put them behind some kind of env var check.

libraries/csharp-net22-runtime/context/Pipe/Program.cs
zeryx: oh yea good point, I'll investigate

libraries/csharp-net22-runtime/context/Pipe/.gitignore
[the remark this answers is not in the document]
kennydaniel: Intellij's gitignore is so so dumb.

You're supposed to check in .idea, but exclude a random subset of files. I think @zeryx has it right.

libraries/csharp-net22-runtime/context/Pipe/Module.cs
zeryx: We're limiting to 1 for simplicity of a PoC, we can have any number, and we can be way more explicit around type checking. We can even have apply's in different modules, but we have to set some kind of structure somewhere. For a basic, fully functional PoC (You can do anything you want with just 1 apply, etc), I think this is ok as is.

WRT the type coercsion issue.. it's still problematic, I'm not sure yet what a solution looks like. I would agree that having a deeper discussion makes sense here, although again outside of scope for a PoC.

libraries/csharp-net22-buildtime/context/test
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+DLL_PATH="$1"
```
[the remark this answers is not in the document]
zeryx: variable and CLI arguments actually aren't necessary if the script is in the pwd for the algorithm project.

zeryx:
In regards to the last commit, I believe the binary support I added works, but if anyone has some example input that I could use for demo purposes here that'd be great. I personally haven't worked with binary input with an algorithm before.
I followed the java algorunner as an example:
https://github.com/algorithmiaio/langpacks/blob/master/java/pipe/src/main/java/com/algorithmia/Pipe.java#L120

libraries/csharp-net22-runtime/context/Pipe/.gitignore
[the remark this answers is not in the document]
zeryx: yup this is the case, closing this.

libraries/csharp-net22-runtime/context/setup
zeryx: How are the `Pipe` artefacts constructed then?

kennydaniel:
@zeryx just post a binary file (like a jpg or something) and write an algorithm that accepts a byte array.

Honestly, your .net binary commit here looks suspicious, why accept an object array...?

zeryx:
Current objectives:
* restructure pipe to be the primary entrypoint for the algorithm at compile time, no more dynamic loading.
*  support binary algorithm input/output.
* add unit tests to ensure compatability of all types, and error message propegation.

languages/csharp-dotnet-2.2/config.json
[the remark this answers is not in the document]
zeryx: Changes made in latest commit.

libraries/csharp-net22-runtime/context/Pipe/Module.cs
zeryx: Refactored into a static compilation system in latest commit.

libraries/csharp-net22-runtime/context/Pipe/Module.cs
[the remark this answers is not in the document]
zeryx: After talking with Kenny, I think we're ok with this only working with one parameter, and one apply.

languages/csharp-dotnet-2.2/config.json
```diff
@@ -0,0 +1,7 @@
+{
+    "artifacts": [
+        {"source":"/opt/algorithm/bin/Release/netcoreapp2.2/__ALGO__.dll", "destination":"/opt/algorithm/pipe.dll"},
+        {"source":"/opt/algorithm/bin/Release/netcoreapp2.2/__ALGO__.runtimeconfig.json", "destination": "/opt/algorithm/pipe.runtimeconfig.json"},
```
zeryx: Turns out with dotnet, the only components required to actually execute a compiled algorithm are the `dll` and `runtimeconfig.json` files, assuming the environment remains the same.

TODO: I need to test this.

languages/csharp-dotnet-2.2/template/pipe/Module.cs
```diff
@@ -0,0 +1,87 @@
...
+            _primaryClassType = typeof(__ALGO__);
+            GetApplyMethod();
+            GetAlgoInputType();
+        }
+        
+
+        private void GetApplyMethod()
+        {
+            // getting only public methods, as those are apply methods
+            MethodInfo applyMethod = _primaryClassType.GetMethod("apply");
```
zeryx: We're only taking the first apply, in most instances this will have zero impact on functionality.

languages/csharp-dotnet-2.2/template/pipe/Module.cs
```diff
@@ -0,0 +1,87 @@
...
+        private void GetAlgoInputType()
+        {
+            ParameterInfo[] parameters = _applyMethod.GetParameters();
+            if (parameters.Length == 0)
+            {
+                throw new Exception("the discovered Apply method for your Algorithm did not have any input arguments!");
+            }
+
+            ParameterInfo paramInfo = parameters[0];
+            _algoInputType = paramInfo.ParameterType;
```
zeryx: To keep things simple, we assume that the apply method has only one required input. This input can be either raw json, or a serializable class defined in __ALGO__.cs

languages/csharp-dotnet-2.2/template_binary/__ALGO__.csproj
```diff
@@ -0,0 +1,15 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
```
zeryx: Obviously template_binary isn't necessary as I was using it just for testing. It could be an alternative template though - the easy decision is to just remove it, thoughts? @kennydaniel

zeryx: Removed.

languages/csharp-dotnet-2.2/template/pipe/Module.cs
zeryx: We also support binary, removed.

zeryx:
Major changes: We migrated the `Pipe` to a nuget package found here: https://www.nuget.org/packages/Algorithmia.Pipe/0.2.0.
The algorithm entrypoint is now a `pipe.cs` file that executes the main pipe wrapper in the nuget package.

languages/csharp-dotnet-2.2/template/__ALGO__.csproj
```diff
@@ -0,0 +1,13 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp2.2</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Algorithmia.Client" Version="(1.0,)" />
+    <PackageReference Include="Algorithmia.Pipe"/>
```
zeryx: Do we want to wildcard the version entirely, or just the major? I'd like to push for a minor/revision version wildcard.

libraries/csharp-net22-runtime/context/pipe
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+dotnet algo.dll
```
zeryx: This should work if the config.json works as I think it does...

zeryx:
Decision decisions:
https://github.com/algorithmiaio/algorithmia-c-sharp-dev - Algorithmia.Pipe lives here
* Due to static compliation issues, we decided to keep the `__ALGO__` string replacement method, rather than attempting to read from the `algorithmia.conf` file at runtime.
* We wildcard the minor/debug versions of `Algorithmia.Pipe` so that new compiles will always fetch the latest backward compatible version.
* As we don't have a `test` system setup yet for the language, I've removed the shell file and entry from the buildtime's dockerfile. I don't think we need to solve this now, and it's something we can add at a later date without adding technical debt.

languages/csharp-dotnet-2.2/config.json
```diff
@@ -0,0 +1,7 @@
+{
+    "artifacts": [
+        {"source":"/opt/algorithm/bin/Release/netcoreapp2.2/__ALGO__.dll", "destination":"/opt/algorithm/algo.dll"},
```
kennydaniel: Question whether this has to be absolute path, or can we remove `/opt/algorithm/` and use relative paths?

libraries/csharp-net22-runtime/context/pipe
kennydaniel: We should time `dotnet run`. If it's like `sbt run` then it takes a while because it's loading the whole dot net cli, instead of directly launching the algo.dll ? Not sure if theres a more direct way to load it for running?

kennydaniel: https://docs.microsoft.com/en-us/dotnet/core/deploying/

kennydaniel: Looks like using `dotnet algo.dll` is the prefered way, and will save space on the framework. Resolving this thread.

zeryx: yea I tried with dotnet run for a while, there is a ton of overhead and it requires literally pushing the entire project into the docker image, rather than just the dll and the configuration files

languages/csharp-dotnet-2.2/config.json
zeryx: @pmcq would love your feedback here, I just copied from the python languages config.json here.

pmcq: i much prefer absolute path but the destination can be anywhere.  the important thing is that there needs to be an entrypoint in `/usr/local/bin/algorithmia-pipe`.  also the `__ALGO__` here isn't going to be templated since this isn't part of the template for an algorithm

libraries/csharp-net22-buildtime/context/build
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+set -e
+
+dotnet restore
```
pmcq: what's restore do? this will be executing in a pretty much clean environment every time, so not sure if that matters

languages/csharp-dotnet-2.2/config.json
zeryx: Oh that's a big deal then. Dotnet creates a dll file with the same name as the incoming project. Would a `/*.dll` work here? Or would that wildcard fail?

pmcq: yeah you should be able to use a wildcard like that  - it just gets translated to a docker copy command here https://github.com/algorithmiaio/langpacks/blob/master/languages/Dockerfile.compile.j2#L10 so anything supported by docker would work

languages/csharp-dotnet-2.2/config.json
```diff
@@ -0,0 +1,7 @@
...
+        {"source":"/opt/algorithm/bin/Release/netcoreapp2.2/__ALGO__.runtimeconfig.json", "destination": "/opt/algorithm/algo.runtimeconfig.json"},
+        {"source":"/opt/algorithm/resources", "destination":"/opt/algorithm/resources"}
```
pmcq: the rust langpack was using this folder name because there's no standard place for resource files to be included in rust, but i bet there is one for c# already (which might need to get bundled in with the build command).  e.g. in scala the best practice is the src/main/resources folder.

languages/csharp-dotnet-2.2/template/__ALGO__.csproj
```diff
@@ -0,0 +1,13 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp2.2</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Algorithmia.Client" Version="(1.0,)" />
```
besirkurtulmus: Can't we make this `1.0.0` instead of `1.0`?

libraries/csharp-net22-buildtime/context/build
zeryx: it downloads the packages from nuget - it's kind of a crappy name but it's like `pip install -r requirements.txt`

languages/csharp-dotnet-2.2/config.json
zeryx: There are no dependencies on the finished system outside of the dll and runtimeconfig.json file, so we could probably just remove that. Kept it for posterity.

languages/csharp-dotnet-2.2/template/__ALGO__.csproj
zeryx: No, it would prevent us from making backwards compatible changes to the client and having it automatically propegate to existing algorithms. Same thing with the Algorithmia.Pipe.

libraries/csharp-net22-runtime/context/Pipe/Program.cs
zeryx: https://stackoverflow.com/questions/26423380/intentionally-cause-a-fatal-exception
https://docs.microsoft.com/en-us/dotnet/api/system.environment.failfast - this is the way of terminating without throwing exceptions. It seems like `Exception` is a sufficient catchall type.
We can certainly revisit this later in the algorithmia-c-sharp-dev package.

languages/csharp-dotnet-2.2/template/src/pipe.cs
```diff
@@ -0,0 +1,16 @@
+using Algorithmia;
+using __ALGO__;
+// This file takes your __ALGO__.cs file, loads the apply function and creates a Pipe wrapper around it.
+// WARNING: Do not remove this file from your project, it will make your algorithm unusable.
+
+namespace __ALGO__
+
+{
+    public class pipe
```
jamesatha: should this be a capital Pipe?

languages/csharp-dotnet-2.2/template/src/pipe.cs
```diff
@@ -0,0 +1,16 @@
...
+    {
+        static int Main()
+        {
+            return AlgorithmiaPipe.Pipe.Enter(typeof(__ALGO__));
```
jamesatha: maybe I need to look at the implementation of the pipe but I thought you would normally pass in a new instance of the `__ALGO__` object

languages/csharp-dotnet-2.2/template/__ALGO__.csproj
```diff
@@ -0,0 +1,13 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp2.2</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Algorithmia.Client" Version="(1.0,)" />
+    <PackageReference Include="Algorithmia.Pipe" Version="(1.0,)" />
```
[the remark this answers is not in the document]
zeryx: Yup!

languages/csharp-dotnet-2.2/template/src/__ALGO__.cs
```diff
@@ -0,0 +1,42 @@
...
+    }
+    
+    class __ALGO__
+    {
+        /// <summary>
+        /// API calls will begin at the apply() method, with the request body passed as 'input'
+        /// For more details, see algorithmia.com/developers/algorithm-development/languages
+        /// </summary>
+        /// <param name="input"></param>
+        /// <returns></returns>
```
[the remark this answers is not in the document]
zeryx: Me neither, honestly they're defaults in rider for /// comments.. I'll just remove them

languages/csharp-dotnet-2.2/template/src/pipe.cs
zeryx: No here we actually want the type, not an object from that class

languages/csharp-dotnet-2.2/template/src/__ALGO__.cs
```diff
@@ -0,0 +1,42 @@
+using System;
+using Algorithmia;
+
+//This is an example of an algorithm, you can evaluate it with this input:
+/// Example Input:
+/// {"name": "Algorithmia user"}
+///
+/// Expected Output:
+/// {"output":"Hello Algorithmia user"}
+namespace __ALGO__
```
anowell: Random: as we look forward to features like forking, I wonder if keeping `__ALGO__` or `__USER__` in the source code (or filenames) for IPA is a good strategy? It might be too late, cuz I imagine the python stuff already does this, but I feel like this is one of those tiny decisions that affects our ability to think about features like forking.

languages/csharp-dotnet-2.2/template/src/pipe.cs
anowell: I was just looking at the magic for `AlgorithmiaPipe`, and while the comment probably belongs in that repo, I don't grok how an algorithm would actually accept binary data. Would the algo itself be responsible for decoding the base64 string?

languages/csharp-dotnet-2.2/template/src/__ALGO__.cs
pmcq: Python does not,  we parse the algorithmia.conf file https://github.com/algorithmiaio/langpacks/blob/master/libraries/python3-runtime/context/pipe#L15

And I also agree with you it's a major decision

languages/csharp-dotnet-2.2/template/src/pipe.cs
zeryx: nope, the pipe deserializes the string into a byte array, you can see it here:
https://github.com/algorithmiaio/algorithmia-c-sharp-dev/blob/master/src/Module.cs#L75

languages/csharp-dotnet-2.2/template/src/__ALGO__.cs
zeryx: As this algorithm is a compiled language, you can't parse the algorithmia.conf file without some kind of runtime reflection. Python is a scripting language which allows you to import dynamically without any design decisions at all. Rust would normally run into this same issue, but I believe with how we do the cargo install, we read from the algorithmia.conf file in advance using a secondary module. I have no idea how to do do this for C#, or whether it's even possible.

We had a conversation about this a few times, specifically dynamic loading of packages using something like a DLL loader (which makes this easy) vs. static compilation (which makes this hard). and it seemed like static compilation won - but maybe there was a misfire in communication somewhere?

What would you like to go folks?

languages/csharp-dotnet-2.2/template/src/pipe.cs
zeryx: It could be? Honestly the class conventions here get weird... I'm not a big fan of capitalizing everything. But just for consistency I can see a case

languages/csharp-dotnet-2.2/template/src/pipe.cs
zeryx: Right now the user is responsible of serializing binary data back into base 64 encoded strings, we could add automatic support for this later - but we don't support that now, so /shrug.

languages/csharp-dotnet-2.2/template/src/__ALGO__.cs
zeryx: We've simplified the problem by removing the file renaming system entirely. From now on we use a generic `Algorithm` namespace, class and filename. This removes issues we may run into with forking and makes things consistent with the Rust client.

languages/csharp-dotnet-2.2/template/src/pipe.cs
zeryx: resolved in latest commit.

languages/csharp-dotnet-2.2/template/__ALGO__.csproj
zeryx: Actually I've updated this doc as the (1.0,) versioning doesn't actually do what I want it to do. From IRC the solution is `1.0.*` instead. No idea why that's not in the .NET docs anywhere...

languages/csharp-dotnet-core2/template/Algorithm.csproj
```diff
@@ -0,0 +1,14 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp2.2</TargetFramework>
```
zeryx: We're actually compatible with any version of netcore 2+ now! but this will be set to 2.2 as it's the latest version, today.

languages/csharp-dotnet-core2/template/src/Algorithm.cs
```diff
@@ -0,0 +1,36 @@
...
+/// {"name": "Algorithmia user"}
+///
+/// Expected Output:
+/// {"output":"Hello Algorithmia user"}
+namespace Algo
+{
+    public class AlgoInput
+    {
+        
+     /// The input object type for your algorithm, this must contain json serializable types.
```
kennydaniel: Seems like this comment should be at the top of the class, not the field?

zeryx: Moved it, along with adding a bit more documentation in the latest commit.

languages/csharp-dotnet-core2/config.json
```diff
@@ -0,0 +1,6 @@
+{
+    "artifacts": [
+        {"source":"/opt/algorithm/bin/Release/*/*.dll", "destination":"/opt/algorithm/Algorithm.dll"},
```
zeryx: the Wildcards need to remain, otherwise if the user changes the name of the primary namespace, we won't migrate the right files over.

languages/csharp-dotnet-core2/template/Algorithm.csproj
```diff
@@ -0,0 +1,16 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <!-- Please ensure that your algorithm only targets a single framework, multiple frameworks are not supported. -->
```
zeryx: This is due to how our config.json file works, I can't imagine a world where this matters, but something to keep in mind.

languages/csharp-dotnet-core2/config.json
jamesatha: Can we hardcode the second `*`  to `Algorithm`  cause we know the class has that name?
This seems strange. are we sure it will target 1 file? if not, then this will throw an error on build cause this has to match one file. Same comment for the next one

languages/csharp-dotnet-core2/template/src/Algorithm.cs
```diff
@@ -0,0 +1,37 @@
...
+//This is an example of an algorithm. You can test it by copy/pasting the following into the web editor:
+/// {"name": "Algorithmia user"}
+///
+/// You should expect to see the following as output:
+/// {"output":"Hello Algorithmia user"}
+
+/// 'Algo' is the primary namespace of your algorithm.
+namespace Algo
+{
+    /// The input object type for your algorithm, the types defined here must be json serializable via the 'Newtonsoft.Json' package.
```
jamesatha: I don't think we want to bother the normal user by talking about the fact that it `must be json serializable via the 'Newtonsoft.Json'`. Even with my C# knowledge, I'm not sure what that means. Maybe say all methods should be public?

languages/csharp-dotnet-core2/config.json
zeryx: We don't know if that class will remain the same, the user can change the names. This resolves IFF there is only one compiled artifact. If we set the gitignore to ignore the `bin` and `obj` directories, we can ensure this is always the case (unless the user messes with the gitignore). I can verify that if there's a single file with that pattern, it works properly.

So the question here is: Do we want to force users to never change the primary namespace, or should we force them to not delete the gitignore and push build products to git (which breaks things)?

The former is simpler for us but restricts the user a bit, the latter might be solvable with some fancy shell scripting

languages/csharp-dotnet-core2/template/src/Algorithm.cs
zeryx: I guess you're right to some degree, I just don't want them to get confused when something doesn't parse right.

zeryx:
> Still a little unsure why mxnet is moving in this PR. Otherwise looks ok to me

I have no idea honestly, I think beacuse when I started this I was on the mxnet branch - but mxnet is merged so :man_shrugging:

languages/csharp-dotnet-core2/config.json
zeryx: I added the build directories to a .gitignore in `template` which should avoid duplicate build products - although I agree it's a little brittle still. I'd be ok with having a discussion around this, but I'd also like to ship something to vistaprint asap and I think this would be a minor pain in the grand scheme...

zeryx: Why don't we merge this, keep it experimental and, if we come up with a better solution eventually, We swap to that before a full rollout?

libraries/csharp-dotnet-core2-buildtime/context/build
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+set -e
+
```
pmcq: fyi - we'll also want the ability to pass in an environment variable to set a mirror for nuget if that's a possbility (i don't know the build tool)

languages/csharp-dotnet-core2/config.json
```diff
@@ -0,0 +1,7 @@
...
+        {"source":"/opt/algorithm/bin/Release/*/*.runtimeconfig.json", "destination": "/opt/algorithm/Algorithm.runtimeconfig.json"},
+        {"source":"/opt/algorithm/resources/", "destination": "/opt/algorithm/resources/"},
```
algorithmia-devops: bad json. get rid of comma at the end of this line

zeryx: whoops, fixed.

libraries/csharp-dotnet-core2-buildtime/context/build
zeryx: oh yea not a bad idea, I'll make sure to add

zeryx: I'll add this in a second QoL PR, closing for now - but I'll comment.

languages/csharp-dotnet-core2/config.json
zeryx: Solutions made - we now move the entire compile directory into `/opt/algorithm`, and run `dotnet *.dll`. This means that we get the best of both worlds, in that there's no brittleness for multiple compiles (it'll still work) - and we can let users change the primary namespace / build products without breaking anything.

languages/Dockerfile.builder.j2
```diff
@@ -20,6 +20,10 @@ RUN /tmp/{{package.script}}
 {% endfor %}
 {% endfor %}
 
+# This is needed for compiled languages, otherwise we get permission errors everywhere
+RUN mkdir /opt/algorithm
+RUN chown $ALGO_UID:$ALGO_UID -R /opt/algorithm
```
pmcq: you don't need -R since you just created an empty directory.  i also think this is specific to c# which ends up creating a `bin` directory specifically whereas something like rust uses `target/release` or whatever.  i think it's okay to make this directory owned by algo but I would just update the comment to say something more specific like `Create the /opt/algorithm directory writeable by algo user so that build scripts can create directories.  For example a compiled language like C# creates a bin directory for the compiled binary`.  I think initially this directory was intentionally not writeable since the build scripts were stored in `/opt/algorithm/bin/build` but since it's moved to `/usr/local/bin/algorithmia-build` i don't think this is a problem

languages/Dockerfile.runner.j2
```diff
@@ -25,6 +25,9 @@ RUN /tmp/{{package.script}}
 {% endfor %}
 {% endfor %}
 
+RUN mkdir /opt/algorithm
+RUN chown $ALGO_UID:$ALGO_UID -R /opt/algorithm
+# Needed for run operations that download/use files in pwd
```
pmcq: What commands are those?  like above I'm not really opposed to this but I don't see anything doing that?  if algorithms need space they should create tmp folders and not use the working directory anyway.

libraries/csharp-dotnet-core2/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+        libncursesw5-dev \
+        libreadline-gplv2-dev \
+        libsqlite3-dev \
+        libssl-dev \
+        tk-dev \
+        uuid-dev \
+        wget && \
+    rm -rf /var/lib/apt/lists/*
+
+RUN apt-get update
```
pmcq: you have a lot of apt-get updates in this dockerfile which are extraneous - the install of apt-transport-https can happen with the rest of the installs above and you should remove the updates on L19 and 25... also where'd you get the list of installs on the first 17 lines? looks like copy/paste from the python installation but those are just from the upstream/default installation of python... https://github.com/docker-library/python/blob/ab8b829cfefdb460ebc17e570332f0479039e918/3.7/stretch/Dockerfile.  I don't see any reason why you need all this other stuff installed.  is there an upstream minimal csharp installation dockerfile you can use instead of this?

languages/Dockerfile.runner.j2
zeryx: This is required for the `dotnet build` operation to function properly, as `algo` needs to own that location to do so.

zeryx: There isn't an easy way to tell the `dotnet` binary to use a different location for build products, it's all kind of hardcoded in unfortunately

libraries/csharp-dotnet-core2/Dockerfile
zeryx: Good question, I'll take a quick look - the update is due to system caching and the fact that there has been new releases to dotnet support since the original `apt-get update`, which causes some issues with getting the latest packages.

languages/Dockerfile.runner.j2
pmcq: this is in the runner image which will never do a `dotnet build`

libraries/csharp-dotnet-core2/Dockerfile
pmcq: what i meant was if you look at lines 1-19 the way to do install apt packages in a dockerfile is usually a big line that does an apt-get update, apt-get install and the removes stuff from /var/lib/apt so that the iamge is as small as possible.  there should not be a standlone apt-get update (and definitely shouldn't have multiple that close together since it means something can likely be combined together)

languages/Dockerfile.runner.j2
zeryx: `dotnet run` I believe requires this still

libraries/csharp-dotnet-core2/Dockerfile
zeryx: There doesn't look like a base image that actually works properly that I can find easily. There probably is, and we can add that in the QoL PR but I don't think this should block merge of this one. I'll remove the extra `update` tho.

languages/Dockerfile.builder.j2
zeryx: Added in latest release

languages/Dockerfile.runner.j2
zeryx: keeping for now, may change in QoL before final release

pmcq:
can you do a squash merge next time. there's a lot of junk commits in git now

libraries/csharp-dotnet-core2/Dockerfile
pmcq: I don't know the framework well but here's microsoft's docs on it-
https://docs.microsoft.com/en-us/dotnet/standard/microservices-architecture/net-core-net-framework-containers/official-net-docker-images
And their image repo on dockerhub - https://hub.docker.com/_/microsoft-dotnet-core

Here's all they have in one image - https://github.com/dotnet/dotnet-docker/blob/master/2.2/runtime/bionic/amd64/Dockerfile#L4-L18

There's still over a dozen libraries in https://github.com/algorithmiaio/langpacks/blob/master/libraries/csharp-dotnet-core2/Dockerfile#L22 that I don't think are necessary.

zeryx:
never heard of a squash merge.. will definitely look to figure out how to do that.

libraries/csharp-dotnet-core2/Dockerfile
zeryx: the microsoft package dotnet-core package doesn't work out of the box, `dotnet` the binary is not available for some reason. Anyways will revisit but certainly not a blocker for alpha testing and deployment for vista.

pmcq:
yeah it's not a big deal just is helpful especially in our repos with a lot commits - instead of clicking the merge button there's an arrow that you can select to change to a squash.  
screenshot from 2019-03-06 08-52-03

libraries/csharp-dotnet-core2/Dockerfile
pmcq: i don't think that we should have users using this in a production sense without fixing this up (and it sounds like you mean deployment for a customer).  At the very least can you create JIRA items for the uncompleted work from this PR so the work is tracked?

zeryx: yeah it's here https://algorithmia.atlassian.net/browse/ALGO-579

zeryx: Becuse of the criticality here I think it's alright as it'll just be in a PoC - (so they can ensure that we do infact support C#), the QoL improvements are definitely on the todo list for this quarterly release, as they effect all other languages as well.

PR #70, opened 2019-02-18
Rust and C++ package sets

anowell:
(disclaimer: low priority, no pressing requirement to CR or merge)

[Feb 2020 Update]
This PR was overhauled Feb 2020 to be more inline with other packageset design decisions. It was successfully validated with the packageset validator script, but hasn't actually been installed into any cluster.


anowell (commented):
2 other comments that don't fit inline anywhere:

- [x] runtime image? I'm not sure what to do here. Rust algos will need libssl (for algo client), but otherwise, it's completely dependent on what libs the algorithm used. In theory, that could be almost anything (e.g. there are even tensorflow bindings in rust), in practice, almost nobody uses rust on our platform. :man_shrugging: 
- [x] C/C++. I've demonstrated that it's possible to build C/C++ code in this package set. In theory, I could create a separate template that stubs out the C/C++ build pieces to make it easier to do, but I'm currently staying away from that because an official decision to support C/C++ deserves more thought than just hacking out a hybrid rust+cpp template.

languages/rust/template/algorithmia.conf
```diff
@@ -0,0 +1,5 @@
+{
+    "username": "__USER__",
+    "algoname": "__ALGO__",
+    "language": "rust"
```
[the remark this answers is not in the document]
zeryx: I think it should be versioned, as even though every release states it's BWC - sometimes there are some minor things (jmalloc?) that aren't and could lead to undefined behavior.

languages/rust/template/src/bin/pipe.rs
```diff
@@ -0,0 +1,165 @@
+// This file provides interop with the langserver
+//
+use algorithm;
```
zeryx: What compiles the `pipe` binary here? Is this compiled along side the primary algorithm project?

languages/rust/template/src/bin/pipe.rs
```diff
@@ -0,0 +1,165 @@
...
+            error: RunnerError {
+                message: error_cause_chain(err),
+                error_type: "SystemError",
+            },
+        }
+    }
+}
+
+fn main() {
+    let mut algo = algorithm::Algo::default();
```
zeryx: Where is this defined? Is this using reflection? I wasn't aware that was possible.

languages/rust/template/src/bin/pipe.rs
```diff
@@ -0,0 +1,165 @@
...
+    println!("PIPE_INIT_COMPLETE");
+    flush_std_pipes();
+
+    let stdin = io::stdin();
+    for line in stdin.lock().lines() {
+        let output_json = match line {
+            Ok(input) => {
+                let output = call_algorithm(&mut algo, input);
```
zeryx: Interesting, so here we're expecting that the downstream algorithm developer will figure out how to parse the arbitrary input object (of which the type is undefined in this instance? I think it's a string?), keeping auto parsing work out of the pipe object itself.

I like this approach honestly, as it keeps things consistent with some of our more successful languages (python, for example) - but I wonder if we'd ever run into any issues with this when attempting to use it for other languages..

zeryx: And I see that you're also not parsing the input json object, `{"metadata":{"content_type":"..."}, "data": "..."}` - are we assuming that the algorithm developer will parse that themselves?
https://github.com/algorithmiaio/langpacks/pull/66/files#diff-f879cfb8ab5d0c3b2f32a894ad0d4d9bR12
https://github.com/algorithmiaio/langpacks/pull/66/files#diff-20313b420f1e2e73ffd5133ef2c6a1bfR32
^ for example in my PR

languages/rust/template/algorithmia.conf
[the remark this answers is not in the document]
anowell: Jmalloc was technically backward compatible. But yes, there are exceptions, usually because of compiler soundness bugs, but as someone who has authored 15+ rust projects over 4 years, never once has a stable compiler upgrade broke my old code; but not upgrading has made it so I couldn't add/update deps.

So as long as there is a straight-forward way to upgrade the compiler for an algorithm every quarter or two, then versioning the rust packageset sounds fine to me

languages/rust/template/src/bin/pipe.rs
anowell: Files in the `src/bin` directory automatically compile as binaries of the same name for a cargo project.

languages/rust/template/src/bin/pipe.rs
anowell: `Cargo.toml` specifies that the lib (`src/lib.rs`) be compiled using the name `algorithm`. This has been the pattern for rust algos since the beginning.

languages/rust/template/src/bin/pipe.rs
anowell: The `call_algorithm` function uses `serde_json`   to deserialize that json object into `struct Request`  and then it turns it into `enum AlgoIo` and then calls `Algo::apply` which is usually generated by the `#[entrypoint]` macro so that the algorithm author can work directly with the type they care about rather than explicitly handling all possit inputs. I admit, this would be a whole lot simpler if not supporting binary input/output.

libraries/rust-buildtime/Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+        libffi-dev \
+        libgdbm-dev \
+        libncursesw5-dev \
+        libreadline-gplv2-dev \
+        libsqlite3-dev \
+        libssl-dev \
+        tk-dev \
+        uuid-dev \
+        wget && \
+    rm -rf /var/lib/apt/lists/*
```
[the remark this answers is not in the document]
pmcq: i'm sure tha tmost of the image was from the ubuntu:16.04 base though? i think that's like 600MB or something

languages/rust/template/algorithmia.conf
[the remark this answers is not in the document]
pmcq: Upgrading the compiler would be:
1. rebuild the rust package set (assuming it's always pulling the latest) 
2. upgrading the version of the packageset in use for an algorithm to the latest build (if you set `packageSetName` when doing a create/update algorithm API call this will happen - we don't have a way yet to show something in the UI like "by the way you can upgrade your base image!")

languages/rust/config.json
```diff
@@ -0,0 +1,6 @@
+{
+    "artifacts": [
+        {"source":"/opt/algorithm/target/release/pipe", "destination":"/usr/local/bin/algorithmia-pipe"},
+        {"source":"/opt/algorithm/resources", "destination":"/opt/algorithm/resources"}
```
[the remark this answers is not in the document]
pmcq: seems like a reasonable location to me - i'm not sure we would need this directory for other languages (except for compiled ones like you mention) and this seems like a good location.  things like java/scala already have standard locations for where they would put resources so I think it's fine to have something that makes sense for rust like this.

pmcq:
>     * [ ]  runtime image? I'm not sure what to do here. Rust algos will need libssl (for algo client), but otherwise, it's completely dependent on what libs the algorithm used. In theory, that could be almost anything (e.g. there are even tensorflow bindings in rust), in practice, almost nobody uses rust on our platform. 🤷‍♂️

My thought is that the base image should be what it is now - small and with the things we _know_ are common.  Beyond that we could be creating packagesets as needed (e.g. rust + ffmpeg bindings).  Ideally we'd have the `+ whatever` extend from the same base image but that's WIP still.

>     * [ ]  C/C++. I've demonstrated that it's possible to build C/C++ code in this package set. In theory, I could create a separate template that stubs out the C/C++ build pieces to make it easier to do, but I'm currently staying away from that because an official decision to support C/C++ deserves more thought than just hacking out a hybrid rust+cpp template.

Not sure I see the C++ stuff in here?  If it was extra work I might rather leave it out from this PR

libraries/rust-buildtime/Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+    && rustc --version \
+    && cargo install --vers 0.2.8 --root /usr/local --no-default-features sccache \
+    && mkdir -p $SCCACHE_DIR \
+    && rm -rf /usr/local/cargo/registry
+
+# Using sccache to force build artifacts to use a cacheable directory
+# Mounting volume for /opt/sccache will allow reusing previously build dependencies
+# Mounting volume for /usr/local/cargo/registry will speed up dependency resolution
+ENV RUSTC_WRAPPER=/usr/local/bin/sccache
+VOLUME ["/usr/local/cargo/registry", "/opt/sccache"]
```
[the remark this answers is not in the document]
pmcq: I don't think that adding the `VOLUME` really adds much here but possibly just leaving the comment in this dockerfile makes sense so we can link to it in JIRA.  There's a general ticket open for enabling package caches.  I'm assuming this stuff needs to be owned by the calling user?

libraries/rust-buildtime/context/test
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+set -e
+
+exec cargo test
```
pmcq: fyi - test doesn't actually function yet 😬 also have a ticket open for that but thanks for adding it in here

libraries/rust-buildtime/context/build
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+set -e
+
+exec cargo build --release --locked
```
pmcq: it doesn't seem like you ever add this or the test script to the buildtime image? e.g. https://github.com/algorithmiaio/langpacks/blob/master/libraries/python2-buildtime/Dockerfile#L1

pmcq (commented):
overall seems like a good strategy but I don't see how an algorithm would actually get built still?  i think just the build script is missing getting copied into the buildtime image. but if that's good then i feel like we should just try out it cause why not... i don't think i can fairly judge without trying to kick the tires of it haha

languages/rust/template/src/bin/pipe.rs
```diff
@@ -0,0 +1,165 @@
+// This file provides interop with the langserver
```
pmcq: one comment on this being in the `template` is that it means algorithm developers can muck with it since it would be in the algorithm's repo (and also it makes it way harder to update across all users).  since this is all statically built i get that the file needs to be available but it could also live in `libraries/rust-buildtime/context/pipe.rs` and copied into the builder image at a well-known location like `/opt/algorithmiaio/rust/pipe.rs` or whatever if there's a way to include that file as part of the build script (or have one line in the build script be `cp /opt/algorithmia/rust/pipe.rs /opt/algorithm/src/pipe.rs` etc).  definitely assists in some ways but would potentially silently overwrite a file with the name pipe.rs if a user actually made that in an algorithm.  idk anything about rust so don't know how hard it is to include sources from multiple locations

libraries/rust-buildtime/context/build
anowell: Fixed in 31dbcdd

libraries/rust-buildtime/context/test
anowell: :+1: - acknowledged as reasonable and resolving this comment.

languages/rust/template/src/bin/pipe.rs
anowell: everything I wrote above is no longer true after the latest update. and the former pipe.rs no longer exists.

languages/rust/template/src/bin/pipe.rs
anowell: this no longer exists after latest commit.

languages/rust/template/src/bin/pipe.rs
anowell: this is not longer accurate after the latest commit. Cargo.toml names the overall project `algorithmia-pipe` now

libraries/rust-buildtime/Dockerfile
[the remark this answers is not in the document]
anowell: Removed in 31dbcdd. Resolving comment.

languages/rust/config.json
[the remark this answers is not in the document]
anowell: Great. I've kept this idea of a resources directory.

libraries/rust-buildtime/Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+
+ENV RUSTUP_HOME=/usr/local/rustup \
+    CARGO_HOME=/usr/local/cargo \
+    PATH=/usr/local/cargo/bin:$PATH \
+    SCCACHE_DIR=/opt/sccache
+
+RUN set -ex \
+    && wget https://static.rust-lang.org/rustup/archive/1.16.0/x86_64-unknown-linux-gnu/rustup-init \
+    && chmod +x rustup-init \
+    && ./rustup-init -y --no-modify-path \
```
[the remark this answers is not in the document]
anowell: @pmcq is the :+1: and ask to hard-code the rustc version in the Dockerfile, or that it's fine to just have this Dockerfile install "latest rustc".

languages/rust/template/src/bin/pipe.rs
anowell: I have removed `pipe.rs`. You will now see the former `lib.rs` became `main.rs` with a very short `fn main()` that registers the apply function. I'd love feedback on this new approach (even open to bikeshedding the precise names, e.g. `setup_handler`)

anowell:
I believe the last commits make this much closer to something we could merge. I believe these are the remaining pieces:

- [ ] Accept or reject the design choice of exposing `fn main` and a function to register the `apply` function. [Pages 9-11 of this doc build the case for this approach](https://docs.google.com/document/d/19HS16SN56uQAxq9vu6Nn_0yvqK9XxemlcwUydpEVmxc/edit#)
- [ ] Nail down versioning story for rust packageset
  - [ ] fixed version or latest stable in Dockerfile?
  - [ ] Should the algorithmia.conf file have some sort of version?
- [ ] Testing - need to re-verify packageset builds, and then actually figure out a good way to test the containerized build/run. The process felt very clunky the last time I tried.
- [ ] Cut a non-beta 3.0 release of the rust client (just waiting on any feedback on the API for registering the handler which has [reference documentation here](https://docs.rs/algorithmia/3.0.0-beta.2/algorithmia/faas/fn.setup_handler.html)) and use it in this template
- [ ] Figure out how to ship this.

languages/rust/template/algorithmia.conf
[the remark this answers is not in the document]
anowell: I interpret that as: I don't have to introduce any sort of version in this algorithmia.conf. Is that correct?

PR #71, opened 2019-02-26
Updating packageset version(s)

besirkurtulmus:
Renaming `allennlp==0.8.1` to `allennlp==0.8` according to PR [here](https://github.com/algorithmiaio/stagetools/pull/130).

Also added the `tensorflow-gpu==2.0` packageset in anticipation.


libraries/tensorflow-gpu-2.0/Dockerfile
```diff
@@ -0,0 +1,2 @@
+RUN pip install keras==2.1.4
```
besirkurtulmus: @zeryx do you remember why we added `keras` to the dockerfile?

templates/tensorflow-gpu-2.0/src/__ALGO__.py
```diff
@@ -0,0 +1,62 @@
+import tensorflow.keras.backend as K
```
besirkurtulmus: I tested this algorithm template in the nightly build on CPU mode, and it worked!

libraries/tensorflow-gpu-2.0/Dockerfile
zeryx: yeah, version control - keras wildcards which version of tensorflow to install which leads to incompatibility.

besirkurtulmus:
@kennydaniel just confirmed that the version range does work as intended in `requirements.txt`

PR #74, opened 2019-03-06
ALGO-579 simplified imports for dotnet sdk installation

zeryx:
reduced imports to bare minimum, since we use ubuntu 16.04 as a base image, the microsoft base images are incompatible - however this method works fine as is.

### DIFF
--- a/libraries/csharp-dotnet-core2/Dockerfile
+++ b/libraries/csharp-dotnet-core2/Dockerfile
@@ -1,4 +1,4 @@
-# Sourced in part from https://github.com/algorithmiaio/langpacks/blob/master/libraries/python36/Dockerfile
+# Sourced in part from https://dotnet.microsoft.com/download/linux-package-manager/ubuntu16-04/sdk-current
 
 RUN apt-get -y update && \
     apt-get install --no-install-recommends -y \
@@ -38,6 +38,7 @@ ENV UBUNTU_VERSION 16.04
 RUN wget -q https://packages.microsoft.com/config/ubuntu/$UBUNTU_VERSION/packages-microsoft-prod.deb
 RUN dpkg -i packages-microsoft-prod.deb
 
+# Installs necessary apt-transport-https package, defined in packages-microsoft-prod.deb
 RUN apt-get install -y apt-transport-https && \
     apt-get update && \
     apt-get install -y dotnet-sdk-$DOTNET_SDK_VERSION
--- a/libraries/csharp-dotnet-core2/Dockerfile
+++ b/libraries/csharp-dotnet-core2/Dockerfile
@@ -41,6 +41,7 @@ RUN dpkg -i packages-microsoft-prod.deb
 # Installs necessary apt-transport-https package, defined in packages-microsoft-prod.deb
 RUN apt-get install -y apt-transport-https && \
     apt-get update && \
-    apt-get install -y dotnet-sdk-$DOTNET_SDK_VERSION
+    apt-get install -y dotnet-sdk-$DOTNET_SDK_VERSION && \
+    rm -rf /var/lib/apt/lists/*
 
 RUN dotnet help
--- a/libraries/csharp-dotnet-core2/Dockerfile
+++ b/libraries/csharp-dotnet-core2/Dockerfile
@@ -1,5 +1,5 @@
 # Sourced in part from https://dotnet.microsoft.com/download/linux-package-manager/ubuntu16-04/sdk-current
-
+# We also used some parts of https://github.com/dotnet/dotnet-docker/blob/master/2.2/sdk/stretch/amd64/Dockerfile
 RUN apt-get -y update && \
     apt-get install --no-install-recommends -y \
         build-essential \
--- a/libraries/csharp-dotnet-core2/Dockerfile
+++ b/libraries/csharp-dotnet-core2/Dockerfile
@@ -2,22 +2,18 @@ RUN apt-get -y update && \
     apt-get install --no-install-recommends -y \
         build-essential \
         ca-certificates \
-        checkinstall \
-        libbz2-dev \
-        libc6-dev \
-        libffi-dev \
-        libgdbm-dev \
-        libncursesw5-dev \
-        libreadline-gplv2-dev \
-        libsqlite3-dev \
         libssl-dev \
-        tk-dev \
-        uuid-dev \
         wget
 
-RUN wget -q https://packages.microsoft.com/config/ubuntu/16.04/packages-microsoft-prod.deb
+# Install .NET Core SDK
+ENV DOTNET_SDK_VERSION 2.2
+ENV UBUNTU_VERSION 16.04
+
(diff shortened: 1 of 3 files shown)


zeryx:
@pmcq I read in the jira ticket that there's an issue with cleaning up /var/apt? Could you elaborate a bit more on that?

pmcq:
you shouldn't be running the apt-get update and installs on separate lines but should instead be following a pattern like this - https://github.com/algorithmiaio/langpacks/blob/master/libraries/python36/Dockerfile#L3-L19 that includes the `rm -rf` at the end

libraries/csharp-dotnet-core2/Dockerfile
```diff
@@ -2,22 +2,18 @@ RUN apt-get -y update && \
...
+
+RUN wget -q https://packages.microsoft.com/config/ubuntu/$UBUNTU_VERSION/packages-microsoft-prod.deb
 RUN dpkg -i packages-microsoft-prod.deb
 
 RUN apt-get install -y apt-transport-https
 RUN apt-get update
-RUN apt-get install -y dotnet-sdk-2.2 
+RUN apt-get install -y dotnet-sdk-$DOTNET_SDK_VERSION
+
+RUN dotnet help
```
pmcq: don't run a help command here

kennydaniel: I think thats a trick to get dotnet to download and build itself. @zeryx ?

zeryx: yeah that actually initializes dotnet, otherwise it'll take a bunch of time during algo execution to conform. It's from the microsoft dotnet sdk dockerfile

zeryx:
> you shouldn't be running the apt-get update and installs on separate lines but should instead be following a pattern like this - https://github.com/algorithmiaio/langpacks/blob/master/libraries/python36/Dockerfile#L3-L19 that includes the `rm -rf` at the end

Unfortunately it's necessary, there is some very odd dependency systems required to get this to build properly for ubuntu 16.04 and that additional update & it being a separate op seems necessary.

libraries/csharp-dotnet-core2/Dockerfile
pmcq: can you include this as a link in the dockerfile and include any license file as necessary then? like https://github.com/algorithmiaio/langpacks/blob/master/libraries/python36/Dockerfile#L1

pmcq:
> > you shouldn't be running the apt-get update and installs on separate lines but should instead be following a pattern like this - https://github.com/algorithmiaio/langpacks/blob/master/libraries/python36/Dockerfile#L3-L19 that includes the `rm -rf` at the end
> 
> Unfortunately it's necessary, there is some very odd dependency systems required to get this to build properly for ubuntu 16.04 and that additional update & it being a separate op seems necessary.

I've literally never seen it being required to be run as separate commands, and refer to the dockerfile best practices guide for some elaboration on why not to do it the way you have it - https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run

libraries/csharp-dotnet-core2/Dockerfile
zeryx: yea definitely can do.

zeryx:
> > > you shouldn't be running the apt-get update and installs on separate lines but should instead be following a pattern like this - https://github.com/algorithmiaio/langpacks/blob/master/libraries/python36/Dockerfile#L3-L19 that includes the `rm -rf` at the end
> > 
> > 
> > Unfortunately it's necessary, there is some very odd dependency systems required to get this to build properly for ubuntu 16.04 and that additional update & it being a separate op seems necessary.
> 
> I've literally never seen it being required to be run as separate commands, and refer to the dockerfile best practices guide for some elaboration on why not to do it the way you have it - https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run

You make some solid points here, I'll do a deeper investigation here tomorrow and see if there is another way to simplify this.

libraries/csharp-dotnet-core2/Dockerfile
zeryx: Should be fixed, what do you think @pmcq ?

pmcq: can you add  `&& rm -rf /var/lib/apt/lists/*` to it too like the first apt-get has

libraries/csharp-dotnet-core2/Dockerfile
```diff
@@ -1,23 +1,46 @@
...
-        libgdbm-dev \
-        libncursesw5-dev \
-        libreadline-gplv2-dev \
-        libsqlite3-dev \
         libssl-dev \
-        tk-dev \
-        uuid-dev \
-        wget
+        wget && \
+        rm -rf /var/lib/apt/lists/*
```
zeryx: I added it here @pmcq

libraries/csharp-dotnet-core2/Dockerfile
zeryx: check above ^

libraries/csharp-dotnet-core2/Dockerfile
pmcq: yeah but you do an apt-get update later which means that it will have contents again

libraries/csharp-dotnet-core2/Dockerfile
```diff
@@ -1,23 +1,46 @@
...
+# Downloads latest dpkg for current ubuntu version
+RUN wget -q https://packages.microsoft.com/config/ubuntu/$UBUNTU_VERSION/packages-microsoft-prod.deb
 RUN dpkg -i packages-microsoft-prod.deb
 
-RUN apt-get install -y apt-transport-https
-RUN apt-get update
-RUN apt-get install -y dotnet-sdk-2.2 
+# Installs necessary apt-transport-https package, defined in packages-microsoft-prod.deb
+RUN apt-get install -y apt-transport-https && \
+    apt-get update && \
```
pmcq: this one

zeryx: Oooh gotcha, should I add it at the last `apt-get install` op, or for each individual command?

pmcq: any apt-get command should look like
``` RUN apt-get update && \
     apt-get install -y <things> && \ 
     rm -rf /var/lib/apt/lists/*
```
see - https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run

zeryx: Done

libraries/csharp-dotnet-core2/Dockerfile
```diff
@@ -1,23 +1,47 @@
...
```
pmcq: you put the `install` before the `update` so this doesn't actually work.  see - https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run.  We also should have tested this again at least in test before merging to prod

PR #76, opened 2019-03-12
[ALGO-551] R 3.4 IPA language support

besirkurtulmus:
This task encompasses getting R 3.4 language support fully operational with our new IPA package build system.

Other relevant PR's:
* https://github.com/algorithmiaio/stagetools/pull/175 / [Compare here](https://github.com/algorithmiaio/stagetools/compare/r-ipa)
* https://github.com/algorithmiaio/common/pull/39 / [Compare here](https://github.com/algorithmiaio/common/compare/r-ipa)
* https://github.com/algorithmiaio/algorithmia/pull/2253 / [Compare here](https://github.com/algorithmiaio/algorithmia/compare/r-ipa)
* https://github.com/algorithmiaio/dockherder/pull/189 / [Compare here](https://github.com/algorithmiaio/dockherder/compare/r-ipa)
* https://github.com/algorithmiaio/web/pull/1151 / [Compare here](https://github.com/algorithmiaio/web/compare/r-ipa)
* algorithmiaio/legit (last thing that needs to be updated: langpack hash version [here](https://github.com/algorithmiaio/legit/blob/master/tools/docker-build-legit#L9))


libraries/r-3.4-buildtime/Dockerfile
```diff
@@ -0,0 +1,9 @@
+COPY --chown=0:0 r-3.4-buildtime/context/build /usr/local/bin/algorithmia-build
+COPY --chown=0:0 r-3.4-buildtime/context/test /usr/local/bin/algorithmia-test
+COPY --chown=0:0 r-3.4-buildtime/context/copy-cran-packages.py /usr/local/bin/algorithmia-copy-cran-packages.py
+COPY --chown=0:0 r-3.4-buildtime/context/rip.py /usr/local/bin/algorithmia-rip.py
+ENV PATH=$PATH:/home/algo/.local/bin
+
+# Install Python2.7 to support the 2 python scripts included in the R builder
+RUN apt-get -y update
+RUN apt-get install -y python2.7
```
kennydaniel: Combine these 2 lines, docker best practices

libraries/r-3.4/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+RUN set -e
+
+ENV DEBIAN_FRONTEND noninteractive
+
+RUN echo "deb http://cran.rstudio.com/bin/linux/ubuntu trusty/" >> /etc/apt/sources.list
+RUN apt-key adv --keyserver keyserver.ubuntu.com --recv-keys E084DAB9
+RUN apt-get -y update
+RUN apt-get install -y r-base=3.4.4-1trusty0
+RUN apt-get install -y r-base-dev=3.4.4-1trusty0
+RUN apt-get install -y libcurl4-gnutls-dev=7.35.0-1ubuntu2.10 # This is needed for RCurl
```
kennydaniel: Combine these apts

libraries/r-3.4-buildtime/Dockerfile
besirkurtulmus: Fixed.

libraries/r-3.4/Dockerfile
besirkurtulmus: Fixed this too in 8f2c67b.

pmcq (commented):
have you been able to test that you can build the docker images for the R build and runtimes?  there's this script here - https://github.com/algorithmiaio/langpacks/blob/master/tools/ipa_packageset_builder.py that you can use to template the dockerfile, and then you would just run a manual `docker build` on that file.  Looks like the script currently only does the runner dockerfile so if you could add a flag so it could template the builder image too that'd be great.  Once we know the images build okay then can also test compiling the algorithm and such as well.  I'm not sure there's a script already that would do that templating but it would also be like.. 20 lines and I can help create that.  That way we can validate things are functioning properly without having to do the whole deploy cycle.

languages/r-3/config.json
```diff
@@ -0,0 +1,7 @@
+{
+    "artifacts": [
+        {"source":"/home/algo/.local", "destination":"/home/algo/.local/"},
```
pmcq: this directory doesn't seem to be used - it's the pip local install directory.

languages/r-3/config.json
```diff
@@ -0,0 +1,7 @@
...
+        {"source":"/opt/algorithm", "destination":"/opt/algorithm/"},
+        {"source":"/home/algo/.Renviron", "destination":"/home/algo/.Renviron"}
```
jamesatha: I don't think we need this. We should be instead copying over the default location that CRAN installs packages

libraries/r-3-runtime/context/pipe
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+set -e
+
+ROOTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd )"
+
+CRAN_MIRROR=${CRAN_MIRROR:=http://cran.rstudio.com/}
+
+echo "options(repos=structure(c(CRAN=\"$CRAN_MIRROR\")))" > ~/.Rprofile
```
jamesatha: You could have added this as an artifact and then we wouldn't need the pipe.r file at all. `pipe` would just be the `pipe.r` file with the right `#!` line at the top to make it execute as r

jamesatha (changes requested):
This seems like as close to a direct port of the existing R stuff as it is now. I don't think that's the approach we want to be taking moving forward. This isn't really leveraging the cool new features that IPA has to offer. I believe this can be simplified and hopefully more R-native if we do that. For starters pipe and pipe.r should be one file. And a lot of the build should be simplified. I think only the rip file needs to be there for parsing the requirements and that's about it

libraries/r-3-buildtime/Dockerfile
```diff
@@ -0,0 +1,9 @@
+COPY --chown=0:0 r-3-buildtime/context/build /usr/local/bin/algorithmia-build
+COPY --chown=0:0 r-3-buildtime/context/test /usr/local/bin/algorithmia-test
+COPY --chown=0:0 r-3-buildtime/context/copy-cran-packages.py /usr/local/bin/algorithmia-copy-cran-packages.py
+COPY --chown=0:0 r-3-buildtime/context/rip.py /usr/local/bin/algorithmia-rip.py
+ENV PATH=$PATH:/home/algo/.local/bin
```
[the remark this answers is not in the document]
besirkurtulmus: I copied this over from the python docker image, and how it's setup. Can remove it, if R doesn't save any info in the `~/.local` directory.

languages/r-3/config.json
besirkurtulmus: Removed in b3bc0d2ba2d7dd09af61f12b1ba84e315b015fd0.

libraries/r-3-buildtime/Dockerfile
```diff
@@ -0,0 +1,9 @@
+COPY --chown=0:0 r-3-buildtime/context/build /usr/local/bin/algorithmia-build
+COPY --chown=0:0 r-3-buildtime/context/test /usr/local/bin/algorithmia-test
+COPY --chown=0:0 r-3-buildtime/context/copy-cran-packages.py /usr/local/bin/algorithmia-copy-cran-packages.py
```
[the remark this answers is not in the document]
zeryx: I believe we still need line 1-2 no?

libraries/r-3-runtime/Dockerfile
```diff
@@ -0,0 +1,3 @@
+COPY --chown=0:0 r-3-runtime/context/pipe /usr/local/bin/algorithmia-pipe
+COPY --chown=0:0 r-3-runtime/context/pipe.r /usr/local/bin/algorithmia-pipe.r
+ENV PATH=$PATH:/home/algo/.local/bin
```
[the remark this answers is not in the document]
zeryx: https://github.com/algorithmiaio/langpacks/blob/master/libraries/python2-runtime/Dockerfile#L1
Are you sure? It looks like we need at least line `1` or `2` right?

besirkurtulmus:
Closing this until IPA lang migration process is clarified.

PR #77, opened 2019-03-28
demo algos should always import Algorithmia

peckjon:


peckjon:
Note to self: location for post-build hash: https://github.com/algorithmiaio/legit/blob/master/tools/docker-build-legit#L9

PR #80, opened 2019-04-30
DS-168 Java Standardization template

zeryx:
This PR defines the user-facing java template code, including the public API for the new AlgorithmHandler.


languages/java8/config.json
```diff
@@ -0,0 +1,7 @@
+{
+    "artifacts": [
+        {"source":"/opt/algorithm/bin/Release/*/*", "destination":"/opt/algorithm/"},
+        {"source":"/opt/algorithm/resources", "destination":"/opt/algorithm/resources/"},
+        {"source":"/home/algo/.nuget", "destination":"/home/algo/.nuget/"}
```
anowell: Isn't nuget for C#, not Java?

languages/java8/template/src/main/java/algorithmia/algorithm/Algorithm.java
```diff
@@ -0,0 +1,26 @@
...
+public class Algorithm {
+    String Apply(String input, HashMap<String, String> context) throws Exception{
+            if(context.containsKey("local_file")){
+                return "hello ".concat(input) + " your model file is downloaded here ".concat(context.get("local_file"));
+            }
+        return "hello ".concat(input);
+    }
+    HashMap<String, String> DownloadModel() throws Exception{
+            HashMap<String, String> context = new HashMap<>();
+            AlgorithmiaClient client = Algorithmia.client(System.getenv("ALGORITHMIA_API_KEY"));
```
anowell: Needs API key? But not Enterprise endpoint?

languages/java8/template/src/main/java/algorithmia/algorithm/Algorithm.java
```diff
@@ -0,0 +1,26 @@
...
+import java.util.HashMap;
+
+public class Algorithm {
+    String Apply(String input, HashMap<String, String> context) throws Exception{
+            if(context.containsKey("local_file")){
+                return "hello ".concat(input) + " your model file is downloaded here ".concat(context.get("local_file"));
+            }
+        return "hello ".concat(input);
+    }
+    HashMap<String, String> DownloadModel() throws Exception{
```
anowell: In general, this download function feels strange for a template. Out of the box, it downloads a file that doesn't exist.

languages/java8/template/src/main/java/algorithmia/algorithm/Algorithm.java
```diff
@@ -0,0 +1,26 @@
...
+import com.algorithmia.algorithmHandler.*;
+import com.algorithmia.*;
+import java.util.HashMap;
+
+public class Algorithm {
+    String Apply(String input, HashMap<String, String> context) throws Exception{
+            if(context.containsKey("local_file")){
+                return "hello ".concat(input) + " your model file is downloaded here ".concat(context.get("local_file"));
+            }
+        return "hello ".concat(input);
```
anowell: Nit: indentation is inconsistent: is this a tabs vs spaces issue?

languages/java8/template/src/main/java/algorithmia/algorithm/AlgorithmBasic.java
```diff
@@ -0,0 +1,16 @@
+package algorithmia;
+import com.algorithmia.algorithmHandler.*;
+import com.algorithmia.*;
+
+class AlgorithmBasic
```
anowell: Why are there two algorithms in the repo template?

languages/java8/template/src/main/java/algorithmia/algorithm/Algorithm.java
```diff
@@ -0,0 +1,26 @@
+package algorithmia;
+import com.algorithmia.algorithmHandler.*;
+import com.algorithmia.*;
+import java.util.HashMap;
+
+public class Algorithm {
+    String Apply(String input, HashMap<String, String> context) throws Exception{
```
anowell: I'm not sure what the prevailing opinion is, but I find myself wanting the template to use json input/output (not merely a string) which most algos end up needing, more than trying to showcase loading (although, maybe we need both?)

languages/java8/template/src/main/java/algorithmia/algorithm/Algorithm.java
pmcq: we shouldn't need either. there is a bug in the algorithmia-java client that doesn't read the ALGORITHMIA_API_KEY environment variable properly but it should be fixed in the java client not here

languages/java8/config.json
zeryx: Yeah I haven't requested feedback yet :)

languages/java8/template/src/main/java/algorithmia/algorithm/Algorithm.java
zeryx: Yes I agree, I'll make an additional PR in the java client while I'm there to get that fixed.

languages/java8/template/src/main/java/algorithmia/algorithm/Algorithm.java
zeryx: You're right, I actually have that file & collection path in the `demo` user, I'll remove the `.my` prefix.

languages/java8/template/src/main/java/algorithmia/algorithm/Algorithm.java
zeryx: It might be, I'll throw it into a linter.

languages/java8/template/src/main/java/algorithmia/algorithm/AlgorithmBasic.java
zeryx: So we can decide as to which one we want to be the primary, and which we want to be secondary. I don't advocate that both should remain in the template when I merge this PR.

languages/java8/template/src/main/java/algorithmia/algorithm/Algorithm.java
zeryx: Agreed, I'd like the advanced variant to do both.

languages/java8/template/src/main/java/algorithmia/algorithm/Algorithm.java
anowell: Is that file present in all enterprise clusters? We will want this to work out-of-box for enterprise customers too.

zeryx: It might make sense to have a demo data file in a users collection, just so we can always show off that api utilization - but that might be scope creep. Would we be fine with using an externally hosted file, like in S3? Or do we want to do away with data API interactions entirely in the advanced template?

zeryx:
Open Question: How do we want to display the template to the user? Obviously we don't want both template versions, but I argue that some of the advanced functionality should be commented out, so it's easy to see and understand.

besirkurtulmus:
> Open Question: How do we want to display the template to the user? Obviously we don't want both template versions, but I argue that some of the advanced functionality should be commented out, so it's easy to see and understand.

Since it's easy to select one – via renaming the files –, I think we should leave it as it is, and decide later when all of the languages are in a workable state.

languages/java8/template/src/main/java/algorithmia/algorithm/Algorithm.java
besirkurtulmus: I always thought that the advanced examples were for internal purposes (for showcasing changes to stakeholders in this project).

We didn't include links to model files in packageset templates, I don't think it makes sense to include them here either. It feels like scope creep.

zeryx: I've adjusted the template to include the advanced functionality in a comment, closing for review.

zeryx:
Closing as I've combined this with my other langpacks PR for simplicity.

PR #81, opened 2019-05-13
DS-181 java 8 package creation

zeryx:
Related PRs:
https://github.com/algorithmiaio/langpacks/pull/80
https://github.com/algorithmiaio/web/pull/1350
https://github.com/algorithmiaio/common/pull/45
https://github.com/algorithmiaio/algorithmia/pull/2338

PR #82, opened 2019-05-17
DS-168 Java Standardization template 

zeryx:
Encompasses the task of creating a Java ojdk 11 IPA packageset & template
template uses handler module found here:
https://github.com/algorithmiaio/algorithm-handler-java
Related PRs:
algorithmiaio/web#1350
algorithmiaio/common#45
algorithmiaio/algorithmia#2338

Created from the merge of: https://github.com/algorithmiaio/langpacks/pull/81 with https://github.com/algorithmiaio/langpacks/pull/80


languages/java8/template/pom.xml
```diff
@@ -0,0 +1,81 @@
...
+            <artifactId>algorithmia-client</artifactId>
+            <version>[1.0.11,)</version>
+        </dependency>
+        <dependency>
+            <groupId>com.algorithmia</groupId>
+            <artifactId>algorithm-handler</artifactId>
+            <version>[1.1,)</version>
+        </dependency>
+        <dependency>
+            <groupId>junit</groupId>
```
kennydaniel: Junit should be marked as a test only dependency

libraries/java8-buildtime/context/build
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
+
+set -e
+
+if [[ -n "$MAVEN_MIRROR" ]] || [[ -n "$JAVA_TOOL_OPTIONS" ]]; then
+	mkdir -p ~/.m2
+	cat > ~/.m2/settings.xml << EOF
+<settings>
+EOF
```
kennydaniel: Does this still work if you put the settings.xml file in the algorithm directory rather than the home directory? Feels awkward overwriting files in the home dir.

languages/java8/template/src/main/java/com/algorithmia/algorithm/Algorithm.java
```diff
@@ -0,0 +1,54 @@
...
+import com.algorithmia.development.*;
+import com.algorithmia.*;
+
+class Algorithm {
+    public String apply(String input) throws RuntimeException {
+        return "hello ".concat(input);
+    }
+
+    public static void main(String[] args) {
+        Algorithm defs = new Algorithm();
```
jamesatha: why is it called `defs`?

languages/java8/template/src/main/java/com/algorithmia/algorithm/Algorithm.java
```diff
@@ -0,0 +1,54 @@
...
+
+
+/**
+ * This class below describes a more advanced template.
+ * If your algorithm needs key data or model files at runtime, you don't want to be downloading that every time your algorithm gets called.
+ * In this situation, you will want to designate a `Loading` function (like DownloadModel below), and provide that method to the AlgorithmHandler constructor.
+ * For more information, please refer to the advanced user guide in https://docs.algorithmia.com
+ */
+
+//public class AlgorithmAdvanced {
```
jamesatha: Just want to make sure, even though this is an Algorithm.java class, the name of the class holding main, can me anything?

languages/java8/template/src/main/java/com/algorithmia/algorithm/Algorithm.java
```diff
@@ -0,0 +1,54 @@
...
+//            return "Hello " + input.name + " you are " + input.age +
+//                    " years old, and your model file is downloaded here " + context.get("local_file");
+//        }
+//        return "hello " + input.name + " you are " + input.age + " years old";
+//    }
+//
+//    HashMap<String, String> downloadModel() throws Exception {
+//        HashMap<String, String> context = new HashMap<>();
+//        AlgorithmiaClient client = Algorithmia.client();
+//        String localFile = client.file("data://demo/collection/testfile.json").getFile().getName();
```
jamesatha: can you replace `demo` with `.my`

languages/java8/template/src/main/java/com/algorithmia/algorithm/Algorithm.java
```diff
@@ -0,0 +1,54 @@
...
+//        context.put("local_file", localFile);
+//        return context;
+//    }
+//
+//    public static void main(String[] args) throws Exception {
+//        AlgorithmAdvanced defs = new AlgorithmAdvanced();
+//        AlgorithmHandler algo = new AlgorithmHandler<>(defs, defs::apply, defs::downloadModel);
```
jamesatha: Are there two types `AlgorithmHandler` (used here) and `Handler` which you use above? I think `AlgorithmHandler` is the better name. And the first arg above is `efs.getClass`. Here you actually pass the instance. I feel like it should be the same thing in both places.

languages/java8/template/src/main/java/com/algorithmia/algorithm/Algorithm.java
```diff
@@ -0,0 +1,54 @@
...
+    public static void main(String[] args) {
+        Algorithm defs = new Algorithm();
+        Handler algo = new Handler<>(defs.getClass(), defs::apply);
+        algo.serve();
+    }
+}
+
+
+/**
+ * This class below describes a more advanced template.
```
jamesatha: I feel like this comment is a little long. Can you just always have an empty load step explaining what to use it for

libraries/java8-buildtime/context/build
```diff
@@ -0,0 +1,57 @@
...
+      <mirrorOf>*</mirrorOf>
+    </mirror>
+  </mirrors>
+
+EOF
+fi
+
+if [[ -n "$HTTPS_PROXY" ]]; then
+        cat >> ~/.m2/settings.xml << EOF
+    <proxies>
```
jamesatha: nit indentation

libraries/java8-buildtime/context/build
jamesatha: there is a chance that you overwrite there setting file in their source. Seems like this is a little less dangerous

jamesatha (commented):
why are we starting with such an old version of java?
Also, I feel like the template needs to be nicer than that but @peckjon / @StephanieKim might have more experience there

languages/java8/template/src/main/java/com/algorithmia/algorithm/Algorithm.java
pmcq: `.my` would be the algorithm caller not the algorithm owner so that might be confusing... that said i'm not sure all the commented out stuff should be in a template.  feels like there should just be a link to a docs page that explains alternatives.

languages/java8/template/src/main/java/com/algorithmia/algorithm/Algorithm.java
zeryx: oh yes good catch, it needs to be the class. will fix.

languages/java8/template/src/main/java/com/algorithmia/algorithm/Algorithm.java
zeryx: Can be renamed, we can also pass `this` if we wanted

languages/java8/template/src/main/java/com/algorithmia/algorithm/Algorithm.java
zeryx: I like that idea honestly, are you comfortable with that style @kennydaniel?

languages/java8/template/src/main/java/com/algorithmia/algorithm/Algorithm.java
zeryx: Yes except they'll need to change the main class in pom.xml as well, otherwise it won't package properly

languages/java8/template/src/main/java/com/algorithmia/algorithm/Algorithm.java
besirkurtulmus: Does `defs` refer to definitions?

besirkurtulmus: Does `defs` stand for definitions?

languages/java8/template/pom.xml
zeryx: There doesn't seem to be a clean way to do this with maven which won't add additional maven/pom complexity, do you have an alternative pom structure in mind?

libraries/java8-buildtime/context/build
zeryx: This is unchanged from langpack build, if you have an example of a better design of this build script lemmie know!

zeryx: There are some templating operations that go on here, we could have a jinja replacement at build time but that's an extra step that langrunner doesn't already do. I would argue that the current approach is relatively sane (although I could see a world where we have it named something other than `settings.xml`)

languages/java11/template/src/main/java/com/algorithmia/algorithm/Algorithm.java
```diff
@@ -0,0 +1,26 @@
...
+
+import com.algorithmia.development.*;
+import com.algorithmia.*;
+
+
+// This class defines your algorithm.
+class Algorithm implements AlgorithmInterface<String, String>{
+
+
+    public String apply(String input){
```
jamesatha: As a first time user, I feel like it's not clear that the input type and output type here needs to match the type parameters of AlgorithmInterface

languages/java11/template/src/main/java/com/algorithmia/algorithm/Algorithm.java
```diff
@@ -0,0 +1,26 @@
...
+        return "hello ".concat(input);
+    }
+
+    // For more information around how loading works, please check out the algorithm docs in algorithmia.com/developers/
+    public void load(){
+        System.out.println("Implement this for loading functionality");
```
jamesatha: I wouldn't print here. Just put a comment

libraries/java11/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+RUN apt-get -y update && \
+    apt-get install --no-install-recommends -y \
+        software-properties-common \
+        build-essential \
+        ca-certificates \
+        libssl-dev \
+        apt-transport-https \
+        wget \
+        default-jdk \
+        openjdk-8-jdk \
```
jamesatha: The rest of the file paths say `java11` so my guess is that this shouldn't be jdk-8 here? Is the default-jdk java11?

libraries/java11-runtime/Dockerfile
```diff
@@ -0,0 +1 @@
+COPY --chown=0:0 java8-runtime/context/pipe /usr/local/bin/algorithmia-pipe
```
jamesatha: so is this java8 or 11?

libraries/java11-buildtime/Dockerfile
```diff
@@ -0,0 +1 @@
+COPY --chown=0:0 java8-buildtime/context/build /usr/local/bin/algorithmia-build
```
jamesatha: here too

jamesatha (commented):
Unsure about the java version we want to use. Also think some documentation on the algorithm code can be made a little better with regards to the generics

languages/java11/template/src/main/java/com/algorithmia/algorithm/Algorithm.java
zeryx: If they don't match they'll get an IDE warning, but we could add a comment here to make that more explicit.

languages/java11/template/src/main/java/com/algorithmia/algorithm/Algorithm.java
zeryx: :+1:

libraries/java11/Dockerfile
zeryx: good point, that will be updated. we want ojdk 11.

libraries/java11-runtime/Dockerfile
zeryx: 11, good catch thanks.

languages/java11/template/src/main/java/com/algorithmia/algorithm/Algorithm.java
besirkurtulmus: They won't get a IDE warning in the web IDE though. I think adding a comment makes sense.

libraries/java8-buildtime/context/build
zeryx: closing due to inactivity

languages/java11/template/src/main/java/com/algorithmia/algorithm/Algorithm.java
```diff
@@ -0,0 +1,35 @@
...
+class Algorithm extends AbstractAlgorithm<String, String>{
+
+    // This class member can be defined as anything you wish, and stores loaded state and mutable state between requests.
+     HashMap<String, String> context = new HashMap<>();
+
+    // This apply function defines the primary motive driver of your algorithm. Please ensure that the types defined in
+    // your algorithm are the same as those defined in your class implementation above ^
+
+    public String apply(String input){
+        // ... = context.get("loaded_model_file")
```
pmcq: i think it would be better practice to just have a member variable `String loadedModelFileName` than to encourage people to store stuff in a hashmap

libraries/java11/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+        ca-certificates \
+        libssl-dev \
+        apt-transport-https \
+        wget \
+        default-jdk \
+        maven && \
+	add-apt-repository ppa:openjdk-r/ppa && \
+	apt-get update && \
+	apt-get install -y \
+	openjdk-11-jdk && \
```
pmcq: this can result in somewhat inconsistent builds since the version of openjdk-11 might change. is it possible to add a version here like `=11.x.x`?  that way every build will be consistent not matter when it is run, and if we want to bump minor versions we'd just edit this file (which results in a different commit in langpacks and future image builds will have that as a label).

i'd try to use a reference dockerfile if possible for someone else who's built reproducible jdk images maybe from https://docs.docker.com/samples/library/openjdk/ (including licenses and stuff)

languages/java11/template/pom.xml
```diff
@@ -0,0 +1,81 @@
...
+    <dependencies>
+        <dependency>
+            <groupId>com.algorithmia</groupId>
+            <artifactId>algorithmia-client</artifactId>
+            <version>[1.0.11,)</version>
+        </dependency>
+        <dependency>
+            <groupId>com.algorithmia</groupId>
+            <artifactId>algorithm-handler</artifactId>
+            <version>1.1.0.rc3</version>
```
pmcq: we would want this to be a real version not a release candidate before publishing

pmcq (commented):
just curious if you've been able to build a sample image with https://github.com/algorithmiaio/langpacks/blob/master/tools/ipa_packageset_builder.py?

languages/java11/template/pom.xml
zeryx: Yes definitely, last change to this PR will be pointing to a real version

languages/java11/template/src/main/java/com/algorithmia/algorithm/Algorithm.java
zeryx: Yeah that's simpler, I can do that

libraries/java11/Dockerfile
zeryx: I'll investigate

languages/java11/template/src/main/java/com/algorithmia/algorithm/Algorithm.java
```diff
@@ -0,0 +1,35 @@
...
+// This class defines your algorithm.
+class Algorithm extends AbstractAlgorithm<String, String>{
+
+    // This class member can be defined as anything you wish, and stores loaded state and mutable state between requests.
+     HashMap<String, String> context = new HashMap<>();
+
+    // This apply function defines the primary motive driver of your algorithm. Please ensure that the types defined in
+    // your algorithm are the same as those defined in your class implementation above ^
+
+    public String apply(String input){
```
anowell: Do we want the basic template to use `String apply(String)`. It seems to me that a very, very small percentage of algorithms use `String` input and output, so the first thing a user has to do is go read the docs to figure out how to use JSON input/output.

Contrast that with my argument around hiding the context and load function from the template: I suspect that the percentage of brand new bare Java algorithms that need context and loading is way lower than the percentage of algos that need to start with some JSON.

languages/java11/template/src/main/java/com/algorithmia/algorithm/Algorithm.java
```diff
@@ -0,0 +1,35 @@
...
+        // ... = context.get("loaded_model_file")
+        return "hello ".concat(input);
+    }
+
+    // For more information around how loading works, please check out the algorithm docs in algorithmia.com/developers/
+    //<--- Implement this for loading functionality -->
+    public void load(){
```
anowell: Same general question: do we really need `load` in the base algorithm template. It seems super useful in a template for "Java with <Framework>" where the template shows how to load a model for that framework. I'm not sure if we have any statistics on what percentage of Java algorithms need a loading step. If it's a low percentage, then I think the extra code and comments are noise, instead of signal.

languages/java11/template/src/main/java/com/algorithmia/algorithm/Algorithm.java
```diff
@@ -0,0 +1,35 @@
...
+import java.util.HashMap;
+
+
+// This class defines your algorithm.
+class Algorithm extends AbstractAlgorithm<String, String>{
+
+    // This class member can be defined as anything you wish, and stores loaded state and mutable state between requests.
+     HashMap<String, String> context = new HashMap<>();
+
+    // This apply function defines the primary motive driver of your algorithm. Please ensure that the types defined in
```
anowell: This language isn't very natural to me. "motive driver". Also "defned in class implementation above" almost made me think of the `HashMap` type that is immediately above this. Can we simplify this language?

As a strawman proposal:

```
// This apply function is the entry point for each API request. 
// The input argument and return types must match the types defined when extending the generic `AbstractAlgorithm`

**Comment by zeryx** [comment]
> Would it be cleaner to use one of the official openjdk docker images as the base image here?
> 
> Something like `openjdk/11.0.3-jdk`
> https://hub.docker.com/_/openjdk/

yup I've done exactly that, thanks for the tip.

File: languages/java11/template/src/main/java/com/algorithmia/algorithm/Algorithm.java [hunk 20]
```diff
@@ -0,0 +1,45 @@
...
+
+
+    // This apply function defines the primary motive driver of your algorithm. Please ensure that the types defined in
+    // your algorithm are the same as those defined in as generic variables in your concrete class defined above.
+
+    public String apply(ExampleInput input){
+        if(input.last_name != null){
+            return "hello " + input.first_name + " " + input.last_name;
+        }
+        else {
```
anowell: Is it standard Java style to put `else` on a separate line from the closing brace of the `if`?

languages/java11/template/src/main/java/com/algorithmia/algorithm/Algorithm.java
```diff
@@ -0,0 +1,45 @@
...
+            return "hello " + input.first_name;
+        }
+    }
+
+
+    public static void main(String[] args) {
+        Algorithm algorithm = new Algorithm();
+        Handler algo = new Handler<>(algorithm);
```
anowell: Curious: this could start another design debate, but what if the |AbstractAlgorithm` had a `final void serve()`. This could be reduced to 

```
Algorithm algo = new Algorithm();
algo.serve();
```

That said, I don't have a strong enough opinion here to push for that, just suggesting it as something I don't recall being discussed.

pmcq (approved):

languages/java11/template/src/main/java/com/algorithmia/algorithm/Algorithm.java
zeryx: This is an interesting idea but something that would cause further divergence from our standard pattern, we can add this functionality later with a backwards compatible change to the algorithm-handler. Merging this for now and continuing testing.

languages/java11/template/src/main/java/com/algorithmia/algorithm/Algorithm.java
jamesatha: @zeryx can we fix this? I'd add the space after the `if` and before the `{`. And else should be on the same line as the other curly

libraries/java11/Dockerfile
```diff
@@ -0,0 +1,72 @@
...
+
+RUN set -eux; \
+	apt-get update; \
+	apt-get install -y --no-install-recommends \
+		bzip2 \
+		unzip \
+		wget \
+		gnupg \
+		gnupg2 \
+     software-properties-common \
```
jamesatha: Looks like you mixed tabs and spaces. Can we make this a little nicer? and I would replace the `;` with `&&`

PR #83, opened 2019-05-22
Support Pytorch 1.1.x

besirkurtulmus:


templates/pytorch-1.1.x/src/__ALGO__.py
```diff
@@ -0,0 +1,56 @@
...
+            if {'matrix_a', 'matrix_b'} <= input_dict.keys():
+                self.A = convert(input_dict['matrix_a'])
+                self.B = convert(input_dict['matrix_b'])
+            else:
+                raise Exception("'matrix_a' and 'matrix_b' must be defined.")
+        else:
+            raise Exception('input must be a json object.')
+        if self.A.shape[-1] != self.B.shape[0]:
+            raise Exception('inner dimensions between A and B must be the same.\n A: {} B: {}'.format(self.A.shape[-1],
+                                                                                                      self.B.shape[0]))
```
jamesatha: nit on formatting. Can we put this on the previous line? or maybe put the 2 args on this line (and indent a bit less

templates/pytorch-1.1.x/src/__ALGO__.py
```diff
@@ -0,0 +1,56 @@
...
+    gpu_tensor = th_tensor.cuda()
+    return gpu_tensor
+
+def apply(input):
+    """
+    Calculates the dot product of two matricies using pytorch, with a cudnn backend.
+    Returns the product as the output.
+    """
+    input = InputObject(input)
+    C = th.mm(input.A, input.B)
```
jamesatha: Doesn't look like the `InputObject` is really needed to be an object

besirkurtulmus: I copied this over from the previous pytorch 1.0.0 template.
I think this is a valid way of parsing inputs outside of the apply() function.

templates/pytorch-1.1.x/src/__ALGO__.py
```diff
@@ -32,8 +32,8 @@ def __init__(self, input_dict):
         else:
             raise Exception('input must be a json object.')
         if self.A.shape[-1] != self.B.shape[0]:
-            raise Exception('inner dimensions between A and B must be the same.\n A: {} B: {}'.format(self.A.shape[-1],
-                                                                                                      self.B.shape[0]))
+            raise Exception('inner dimensions between A and B must be the same.\n A: {} B: {}'\
+                .format(self.A.shape[-1], self.B.shape[0]))
```
besirkurtulmus: @jamesatha does this look better?

templates/pytorch-1.1.x/src/__ALGO__.py
besirkurtulmus: Fixed it.

besirkurtulmus:
@zeryx Yes, I have confirmed it.

templates/pytorch-1.0.0/src/__ALGO__.py
```diff
@@ -32,8 +32,8 @@ def __init__(self, input_dict):
         else:
             raise Exception('input must be a json object.')
         if self.A.shape[-1] != self.B.shape[0]:
-            raise Exception('inner dimensions between A and B must be the same.\n A: {} B: {}'.format(self.A.shape[-1],
-                                                                                                      self.B.shape[0]))
+            raise Exception('inner dimensions between A and B must be the same.\n A: {} B: {}'\
```
jamesatha: nit. keep the `.format(` on the previous line so you don't need the slash

templates/pytorch-1.1.x/src/__ALGO__.py
jamesatha: I'm not a fan of the slash. I'd move the method to the previous line and have the args on the second line. But don't feel too strongly

templates/pytorch-1.1.x/src/__ALGO__.py
jamesatha: I guess for consistencies sake, we can keep this as is, but it seems like a method that does the verification of these 2 dimensions would be nicer. But not  going not going to block this PR for that

templates/pytorch-1.1.x/src/__ALGO___test.py
```diff
@@ -0,0 +1,8 @@
+from __ALGO__ import apply
+
+def test_algorithm():
+    input = {"matrix_a": [[0, 1], [1, 0]], "matrix_b": [[25, 25], [11, 11]]}
+    result = apply(input)
+    assert result == {"product": [[11., 11.], [25., 25.]]}
```
jamesatha: do we really need the dots? 11 and `11.` are going to be equal always, right/

besirkurtulmus: I think the results are floats, hence why we need the dots.

besirkurtulmus: Also, I think we want to be explicit about it.

PR #84, opened 2019-05-31
DS-171 template & langpacks work for R IPA migration

zeryx:
this PR describes the new R template, which includes a `load` function and the concept of an `algorithmHandler`.
This PR also encompasses all libraries & languages work required for the support of the r34 language packageset.


languages/r34/template/src/Algorithm.r
```diff
@@ -0,0 +1,15 @@
+library(algorithmia)
+# API calls will begin at the algorithm() method, with the request body passed as 'input'
+# For more details, see algorithmia.com/developers/algorithm-development/languages
+
+algorithm <- function(input) {
+  paste("hello", input)
+}
+
+load <- function() {
+  # Defines the loading function if defined
```
anowell: Nit: "define ... if defined" reads a bit awkward.. perhaps just "Optionally define..."

languages/r34/template/src/Algorithm.r
```diff
@@ -0,0 +1,15 @@
...
+}
+
+algo <- getAlgorithmHandler(algorithm, load)
+algo$run()
```
anowell: Didn't you use `serve` instead of `run` in the final Java implementation (or am I mistaken)?

zeryx: oh yes good catch will rename

languages/r34/template/src/Algorithm.r
zeryx: yea good call

languages/r36/template/algorithmia.conf
```diff
@@ -0,0 +1,5 @@
+{
+    "username": "__USER__",
+    "algoname": "__ALGO__",
+    "langauge": "r36"
```
StephanieKim: misspelling of "language" says "langauge" - prob someone already caught that tho

languages/r36/template/src/Algorithm.r
```diff
@@ -0,0 +1,36 @@
...
+  # model_file_path <- client$file('data://path/to/my/modelFile.hd5)$getFile()$name
+  # keras_model <- keras.load_model(model_path)
+  # keras_model
+}
+
+# This code turns your library code into an algorithm that can run on the platform.
+# If you intend to use loading operations, remember to pass a `load` function as a second variable.
+algo <- getAlgorithmHandler(algorithm, load)
+# The 'serve()' function actually starts the algorithm, you can follow along in the source code.
+algo$serve()
```
StephanieKim: This is dope thanks for adding all the comments!

stevielkim (changes requested):
Just the misspelling of "language" that I commented on all else is good

stevielkim (approved):
Nothing stands out as a blocker, but would nudge you away from using inline comments and encourage you to use docstrings for functions/modules and above line comments for single line comments.

languages/r36/template/bin/copy-cran-packages.py
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/python
```
jamesatha: we don't want a bin directory anymore
Also, I don't think we need the copy script anymore (more in a follow up comment)

libraries/r36-buildtime/Dockerfile
```diff
@@ -0,0 +1 @@
+COPY --chown=0:0 r36-buildtime/context/build /usr/local/bin/algorithmia-build
```
jamesatha: You want to be copying over the other parts of what are currently in the `bin` dir into here so it's not in the source repos

libraries/r36-buildtime/context/build
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
```
jamesatha: I think this can be simplified a bit. We don't have the shared cache right now and we can install dependencies in the normal location so I think we don't have to do the copying and can just make 1 call to rip

languages/r36/template/src/Algorithm.r
```diff
@@ -0,0 +1,36 @@
...
+#   algorithm <- function(input, keras_model) {
+#   ....
+#   prediction <- keras_model.predict(input)
+#   result <- list(class <- prediction[0], confidence <- prediction[1])
+#   result
+# }
+
+
+load <- function() {
+  # Here you can optionally define a function called when the algorithm is loaded.
```
jamesatha: do you want to be explicit and return null?

languages/r36/template/src/Algorithm.r
```diff
@@ -0,0 +1,36 @@
+library(algorithmia)
+# API calls will begin at the algorithm() method, with the request body passed as 'input'
+# For more details, see algorithmia.com/developers/algorithm-development/languages
+
+algorithm <- function(input) {
+  paste("hello", input)
+}
+
+# Here is an example of an advanced form of an algorithm function, using a load function (like the example below)
```
jamesatha: there are a ton of comments in here. To me it seems a bit more complicated this way

jamesatha (changes requested):

languages/r36/template/bin/copy-cran-packages.py
zeryx: yea Ill throw all of this into the context

libraries/r36-buildtime/context/build
zeryx: for sure i just wanted to make this PR a refactor for the pieces that didn't need to change

languages/r36/template/src/Algorithm.r
zeryx: sure I can add that

languages/r36/template/src/Algorithm.r
zeryx: was a request by @StephanieKim, I'm indifferent haha

languages/r36/config.json
```diff
@@ -0,0 +1,7 @@
+{
+    "artifacts": [
+        {"source":"/opt/algorithm", "destination":"/opt/algorithm/"},
+	{"source":"/home/algo/.Renviron", "destination":"/home/algo/.Renviron"}
```
jamesatha: nit: indentation

libraries/r36-buildtime/context/build
jamesatha: The point of migrating to IPA is to get a chance to fix/improve the way we do things. Bringing over this old cruft to the master branch doesn't seem right to me. Having to install things once should also reduce build times drastically for R

libraries/r36/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+		git \
+		libcurl4-gnutls-dev \
+		wget
+
+RUN Rscript -e "install.packages('rjson')" \
+    && Rscript -e "install.packages('base64enc')" \
+    && Rscript -e "install.packages('RCurl')" \
+    && Rscript -e "install.packages('pacman')"
+
+# We want pacman, rjson, and base64enc to be in the docker image. The others can be bind mounted
```
jamesatha: I don't think we need anything from this line until the `RUN chown algo -R /usr/local/lib/R/site-library` (we want the algo user to be able to install to site-library directory)

jamesatha (commented):
Don't feel happy merging this into master. I feel like it's got a lot of the old `classic` hacks and we should be able to be a bit more normal (in terms of R) with IPA world

libraries/r36/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+		r-base-dev \
+		libssl-dev \
+		libxml2-dev \
+		python-dev \
+                apt-transport-https \
+		ca-certificates \
+		git \
+		libcurl4-gnutls-dev \
+		wget
+
```
jamesatha: Something strange that I hadn't noticed is that we are just using the default CRAN repo and we don't need this: https://github.com/algorithmiaio/langpacks/blob/master/r/template/bin/setup#L15

libraries/r36/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+RUN apt-get update \
+	&& apt-get install -y --no-install-recommends \
+		build-essential \
+		software-properties-common \
+		r-base-dev \
+		libssl-dev \
+		libxml2-dev \
+		python-dev \
+                apt-transport-https \
```
jamesatha: nit spacing

jamesatha (approved):
ship it!

PR #85, opened 2019-05-31
Caffe GPU + Python2.7 Packageset

field-cady:
This has dockerfiles for caffe on Python27 and Python37.
I need to fix the GPU support for python 37


libraries/caffe_python37/Dockerfile
```diff
@@ -0,0 +1,156 @@
+# Lightly modified from https://github.com/docker-library/python/blob/ab8b829cfefdb460ebc17e570332f0479039e918/3.7/stretch/Dockerfile
+#FROM ubuntu:18.04
+FROM nvidia/cuda
```
pmcq: worth talking to besir about this but these shouldn't be full dockerfiles like this but just snippets that are relevant for caffe specifically, we glob all the snippets together for a given package set

besirkurtulmus: We've already talked about it. He's going to delete this from his PR.

besirkurtulmus:
Closing PR because caffe got dropped.

PR #87, opened 2019-06-04
DS-181 adding backup keyserver if main fails.

zeryx:
https://github.com/algorithmiaio/langpacks/commit/63a4556dc0f1d42bea4ee7cb9e839dc01f84ae1c

### DIFF
--- a/libraries/java11/Dockerfile
+++ b/libraries/java11/Dockerfile
@@ -49,9 +49,9 @@ RUN set -eux; \
 	\
 	export GNUPGHOME="$(mktemp -d)"; \
 # TODO find a good link for users to verify this key is right (https://mail.openjdk.java.net/pipermail/jdk-updates-dev/2019-April/000951.html is one of the only mentions of it I can find); perhaps a note added to https://adoptopenjdk.net/upstream.html would make sense?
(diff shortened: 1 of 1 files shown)


libraries/java11/Dockerfile
```diff
@@ -49,9 +49,9 @@ RUN set -eux; \
 	\
 	export GNUPGHOME="$(mktemp -d)"; \
 # TODO find a good link for users to verify this key is right (https://mail.openjdk.java.net/pipermail/jdk-updates-dev/2019-April/000951.html is one of the only mentions of it I can find); perhaps a note added to https://adoptopenjdk.net/upstream.html would make sense?
-	gpg --batch --keyserver ha.pool.sks-keyservers.net --recv-keys CA5F11C6CE22644D42C6AC4492EF8D39DC13168F; \
+	$(gpg --batch --keyserver ha.pool.sks-keyservers.net --recv-keys CA5F11C6CE22644D42C6AC4492EF8D39DC13168F  ||  gpg --batch --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys "$GPG_KEY"); \
```
pmcq: you never set `GPG_KEY` so the backup won't actually do anything here - here's the reference from python  - https://github.com/algorithmiaio/langpacks/blob/master/libraries/python27/Dockerfile#L49-L58

PR #89, opened 2019-06-17
DS-276 Create langpack packagset validator script

zeryx:
This script allows a user to validate that a new IPA package(set) will function as expected on our platform.
Specifically this allows us to test the langserver -> pipe -> algorithm comunication pipeline to verify that new algorithms, and new packages work properly.
It also is useful for verifying that our docker images can build and run effectively.

Things this doesn't support:

- GPU access (we don't use nvidia-docker.. yet)
- Custom template locations (must be in templates/ or languages/)
- Non-standard docker configurations (pulling from remote services, etc)


besirkurtulmus:
Can we also have a basic README explaining what it does? :)

languages/Dockerfile.compileLocal.j2
```diff
@@ -0,0 +1,15 @@
+FROM {{builder_image}} as builder
+USER 2222
+# Docker build commands don't resolve environment variables so need this to either be numeric or a build argument
+COPY --chown=2222:2222 algosource /opt/algorithm/
+COPY --chown=2222:2222 {{config.src_path}} {{config.dest_path}}
```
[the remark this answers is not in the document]
zeryx: is it possible to "ignore" the line in the event that we don't want to migrate local packages with jenkins? Would `config.src_path = None` and `config.dest_path = None` accomplish that? Maybe an if statement would work here. 
I'll ping you on slack for ideas.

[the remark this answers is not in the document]
pmcq: yeah i'm not really sure why there is a separate dockerfile here for local stuff? i would strongly prefer to use the same dockerfile that gets used by the rest of the system or they will likely drift and we'll end up with errors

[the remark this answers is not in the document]
zeryx: Solved, replaced with an if check in compile.j2

languages/Dockerfile.compile.j2
```diff
@@ -2,6 +2,11 @@ FROM {{builder_image}} as builder
 USER 2222
 # Docker build commands don't resolve environment variables so need this to either be numeric or a build argument
 COPY --chown=2222:2222 algosource /opt/algorithm/
+
+{% if config.src_path is defined %}
```
pmcq: i mean i guess i still have the question as to what the contents of this directory are and why it is added for local compilation.  scanning code i'm not really certain but maybe it's something with trying to avoid installing dependencies on every local build or something, but i would really rather not have this in here even though it can add compile time for local development.  we have a ticket # 2469 for getting dependency caches from the host to work properly and i would rather have it work for both local and cluster builds rather than have an extra parameter that isn't used otherwise

zeryx: I added it to make my life easier for when deploying a required dependency to a third party repository was expensive to resolve (maven, CRAN, etc), so adding this allows you to migrate a local package to the docker image for testing. If you are planning on enabling that through a dedicated host -> container dependency pipeline I can definitely remove it or put it in a separate PR.

zeryx: Renamed the variable to be a bit more articulate

tools/template_manager.py
```diff
@@ -1,14 +1,17 @@
-#!/usr/bin/env python3
 from jinja2 import Template
-import argparse
 from os import path
 from os.path import isfile
+
```
StephanieKim: Needs docstrings please :-)

stevielkim (commented):
Looks good - couple of small nitpicky changes along with also adding docstrings for builder script :-)

tools/template_manager.py
zeryx: I won't be adding any to this other file, this is an internal API that should never need to be interacted with specifically during normal validation procedures.

PR #91, opened 2019-06-17
dockerfile working for phantomjs & chrome

field-cady:
Adding a dockerfile that can be used for an image that runs the selenium library


besirkurtulmus:
Can we remove the `FROM`  section & the Python stuff?

They're added in the backend anyway.

field-cady:
Yeah they were mostly just there for completeness of reproducibility

pmcq:
theres a testing script james made that you can use to test, can make sure to send that your way tomorrow

libraries/selenium-python37/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+  wget https://chromedriver.storage.googleapis.com/2.41/chromedriver_linux64.zip && \
+  unzip chromedriver_linux64.zip && \
+  cp chromedriver /usr/local/share/ && \
+  ln -s /usr/local/share/chromedriver /usr/local/bin/ && \
+  DEBIAN_FRONTEND=noninteractive apt-get install --no-install-recommends -y  chromium-browser
+
+RUN pip install selenium
+RUN pip install Algorithmia
+
+# python -c "import selenium;from selenium import webdriver;driver = webdriver.PhantomJS()"
```
pmcq: remove?

libraries/selenium-python37/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+  ln -s /usr/local/share/phantomjs-2.1.1-linux-x86_64/bin/phantomjs /usr/local/bin/
+
+# Chrome for Selenium
+RUN apt-get update && \
+  apt-get install unzip && \
+  wget https://chromedriver.storage.googleapis.com/2.41/chromedriver_linux64.zip && \
+  unzip chromedriver_linux64.zip && \
+  cp chromedriver /usr/local/share/ && \
+  ln -s /usr/local/share/chromedriver /usr/local/bin/ && \
+  DEBIAN_FRONTEND=noninteractive apt-get install --no-install-recommends -y  chromium-browser
```
pmcq: at end of this you should add a `&& rm -rf /var/lib/apt/lists/*` (see notes on apt-get installs here: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run)

libraries/selenium-python37/Dockerfile
field-cady: I am down with removing the Algorithmia line.  But I feel like we should keep selenium - the previous parts of the Dockerfile are are only meaningful if they are followed up by pip-installing the selenium library, and I feel like we should have the Dockerfile be "all or nothing" in installing selenium.  We already do pip-install in the Dockerfile, like with the AllenNLP language pack.

pmcq:
also fyi - the script for validating the packagesets locally is here - https://github.com/algorithmiaio/langpacks/blob/master/tools/ipa_packageset_builder.py

templates/selenium-python37/src/__ALGO___test.py
```diff
@@ -0,0 +1,8 @@
+from __ALGO__ import apply
+
+def test_algorithm():
+    input = {}
+    result = apply(input)
+    assert result == {"status": "success"}
+
+test_algorithm()
```
jamesatha: You shouldn't need this line. normal pytest will run all methods that start with "test"
Example: https://github.com/algorithmiaio/langpacks/blob/master/python/template/src/__ALGO___test.py

(Also, testing in IPA doesn't work... just saying)

templates/selenium-python37/src/__ALGO__.py
```diff
@@ -0,0 +1,31 @@
...
+    driver = webdriver.PhantomJS()
+    return driver
+
+# API calls will begin at the apply() method, with the request body passed as 'input'
+# For more details, see algorithmia.com/developers/algorithm-development/languages
+def apply(input):
+    driver = get_chrome()
+    #driver = get_phantomjs()
+    driver.get("http://www.python.org")
+    assert "Python" in driver.title
```
jamesatha: Don't think we should have asserts in the apply method

templates/selenium-python37/src/__ALGO__.py
```diff
@@ -0,0 +1,31 @@
...
+
+def get_phantomjs():
+    driver = webdriver.PhantomJS()
+    return driver
+
+# API calls will begin at the apply() method, with the request body passed as 'input'
+# For more details, see algorithmia.com/developers/algorithm-development/languages
+def apply(input):
+    driver = get_chrome()
+    #driver = get_phantomjs()
```
jamesatha: is this an alternative? Can we document it instead of just having it like this. or remove it

templates/selenium-python37/src/__ALGO__.py
```diff
@@ -0,0 +1,31 @@
...
+    driver.get("http://www.python.org")
```
jamesatha: If the user chooses to not give the algorithm internet permission, this is going to barf, right? Is there a way to make a default that doesn't need internet permissions?

PR #92, opened 2019-06-19
Selenium template

field-cady:
improving the template code


pmcq:
actually i had a question about package naming in another PR - can talk over there though

templates/selenium3.141.x-python37/src/__ALGO__.py
```diff
@@ -12,20 +13,26 @@ def get_chrome():
...
+    # PhantomJS must put logs in a place where it has write permissions
+    driver = webdriver.PhantomJS(service_log_path="/tmp/ghostdriver.log")
     return driver
 
 # API calls will begin at the apply() method, with the request body passed as 'input'
 # For more details, see algorithmia.com/developers/algorithm-development/languages
 def apply(input):
+    # note: you MUST have internet access enabled for this test work
     driver = get_chrome()
     #driver = get_phantomjs()
```
jamesatha: Still not a fan of the commented out line. Why not have a `getDriver` method that takes in an enum that is either phantomjs or chrome?

templates/selenium3.141.x-python37/src/__ALGO__.py
```diff
@@ -12,20 +13,26 @@ def get_chrome():
...
     driver.get("http://www.python.org")
-    assert "Python" in driver.title
+    if 'Python' not in driver.title:
+        return {"status": "failure",
```
jamesatha: I'd put this whole object on 1 line. same below

templates/selenium3.141.x-python37/src/__ALGO__.py
```diff
@@ -12,20 +13,26 @@ def get_chrome():
...
 def get_phantomjs():
-    driver = webdriver.PhantomJS()
+    # PhantomJS must put logs in a place where it has write permissions
+    driver = webdriver.PhantomJS(service_log_path="/tmp/ghostdriver.log")
     return driver
 
 # API calls will begin at the apply() method, with the request body passed as 'input'
 # For more details, see algorithmia.com/developers/algorithm-development/languages
 def apply(input):
+    # note: you MUST have internet access enabled for this test work
```
jamesatha: I'm still not a huge fan of this. I would prefer this wasn't the case. Maybe we can add something to jira about making package sets that need internet so at least the default internet permission would be correct. @kennydaniel might have thoughts on that.

An alternative would be to spin up a local server and use that but that might look gross and not too useful

templates/selenium3.141.x-python37/src/__ALGO__.py
```diff
@@ -12,20 +13,26 @@ def get_chrome():
...
     elem = driver.find_element_by_name("q")
     elem.clear()
     elem.send_keys("pycon")
     elem.send_keys(Keys.RETURN)
-    assert "No results found." not in driver.page_source
+    if 'No results found' in driver.page_source:
+        return {"status": "failure",
+            "message": "Failed to find required text in page body"}
     driver.close()
     return {"status": "success"}
```
jamesatha: I feel like this would be more useful if this returned the title and body of the page cause right now you aren't really doing much here

templates/selenium3.141.x-python37/src/__ALGO___test.py
```diff
@@ -4,5 +4,3 @@ def test_algorithm():
     input = {}
```
jamesatha: This seems like a useless test. I think it would be cooler if you passed the in the url, find_by_element and input as parameters and did the checks that you have in the apply method here instead. And if you are returning the title/body in apply, that will be easy

templates/selenium3.141.x-python37/src/__ALGO__.py
besirkurtulmus: These checks are based on the official selenium examples.

We wanted to we in sync w/ that as much as possible.

PR #96, opened 2019-07-22
DS-218 added IPA anaconda3.5 algorithm support

zeryx:
* Uses the same algorithmia-python pipe logic that vanilla python 2, 3 will use in revised versions.
* Uses the yaml file standard for use with conda create
* Can use multiple different versions of python without needing custom IPA images
* Can use anaconda & pypi specific packages

### DIFF
--- a/languages/anaconda3/config.json
+++ b/languages/anaconda3/config.json
@@ -1,6 +1,7 @@
 {
     "artifacts": [
-        {"source":"/home/algo/.local", "destination":"/home/algo/.local/"},
+        {"source":"/home/algo/.conda", "destination":"/home/algo/.conda/"},
+	    {"source":"/home/algo/.cache", "destination":"/home/algo/.cache/"},
         {"source":"/opt/algorithm", "destination":"/opt/algorithm/"}
     ]
 }
--- a/languages/anaconda3/config.json
+++ b/languages/anaconda3/config.json
@@ -1,7 +1,7 @@
 {
     "artifacts": [
         {"source":"/home/algo/.conda", "destination":"/home/algo/.conda/"},
-	    {"source":"/home/algo/.cache", "destination":"/home/algo/.cache/"},
+	{"source":"/home/algo/.cache", "destination":"/home/algo/.cache/"},
         {"source":"/opt/algorithm", "destination":"/opt/algorithm/"}
     ]
 }
--- a/languages/anaconda3/config.json
+++ b/languages/anaconda3/config.json
@@ -1,0 +1,6 @@
+{
+    "artifacts": [
+        {"source":"/home/algo/.local", "destination":"/home/algo/.local/"},
+        {"source":"/opt/algorithm", "destination":"/opt/algorithm/"}
+    ]
+}
--- a/languages/anaconda3/template/requirements.yaml
+++ b/languages/anaconda3/template/requirements.yaml
@@ -3,9 +3,8 @@ channels:
 - conda-forge
 - defaults
 dependencies:
-# Here are your anaconda defined dependencies, including the python version
-# If you wish to use a different python version, please adjust the value below
-- python=3.7
+# Here are where your anaconda dependencies can be defined
+- python
 # If you wish to use a specific version of pip, please adjust the value below
 - pip=19.0
 - pip:
--- a/languages/anaconda3/template/requirements.yaml
+++ b/languages/anaconda3/template/requirements.yaml
@@ -1,0 +1,13 @@
+channels:
+# Add your desired anaconda channels here
+  - conda-forge
+  - defaults
+dependencies:
+# Here are your anaconda defined dependencies, including the python version
+# If you wish to use a different python version, please adjust the value below
+  - python=3.6
+# If you wish to use a specific version of pip, please adjust the value below
+  - pip=19.0
+  - pip:
+# Add your pip dependencies (taken directly from pypi) here, we'll use the version of python and pip installed above
+    algorithmia>=1.1.4
--- a/languages/anaconda3/template/requirements.yaml
+++ b/languages/anaconda3/template/requirements.yaml
@@ -1,14 +1,14 @@
 channels:
 # Add your desired anaconda channels here
-  - conda-forge
-  - defaults
+- conda-forge
+- defaults
 dependencies:
 # Here are your anaconda defined dependencies, including the python version
 # If you wish to use a different python version, please adjust the value below
-  - python=3.7
+- python=3.7
 # If you wish to use a specific version of pip, please adjust the value below
-  - pip=19.0
-  - pip:
+- pip=19.0
+- pip:
 # Add your pip dependencies (taken directly from pypi) here, we'll use the version of python and pip installed above
-    algorithmia>=1.2,<1.3
-    six>=1.12.0
+  - six>=1.12.0
+  - algorithmia>=1.2,<1.3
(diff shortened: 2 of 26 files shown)


languages/anaconda3/template/requirements.yaml
```diff
@@ -0,0 +1,14 @@
+channels:
+# Add your desired anaconda channels here
+- conda-forge
+- defaults
+dependencies:
+# Here are your anaconda defined dependencies, including the python version
+# If you wish to use a different python version, please adjust the value below
```
pmcq: doesn't that mean that every algorithm would end up with their own installation of python then? that feels pretty inefficient

zeryx: yes, I believe we can pre-bake a version into the docker image and use that instead if pre-installed though. But this does allow us to have 1 docker image for _any_ python version which I think is pretty versatile!

zeryx: We could also skip that line and use the preinstalled version of python in the image (aka using a python37 base image or w/e along with the anaconda image) and just not provide it as a default customization option.

libraries/anaconda35/Dockerfile
```diff
@@ -0,0 +1,25 @@
+#  $ docker build . -t continuumio/anaconda3:latest -t continuumio/anaconda3:5.3.0
+#  $ docker run --rm -it continuumio/anaconda3:latest /bin/bash
+#  $ docker push continuumio/anaconda3:latest
+#  $ docker push continuumio/anaconda3:5.3.0
+
+ENV LANG=C.UTF-8 LC_ALL=C.UTF-8
+ENV PATH /opt/conda/bin:$PATH
+
+RUN apt-get update --fix-missing && apt-get install -y wget bzip2 ca-certificates \
```
dmorris-algo: Any reason we can't move the `curl`, `grep`, `sed`, and `dpkg` packages up here and then follow the [`rm pattern`](https://github.com/algorithmiaio/langpacks/blob/master/libraries/java11/Dockerfile#L21).

libraries/anaconda35/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+    libglib2.0-0 libxext6 libsm6 libxrender1 \
+    git 
+
+RUN wget --quiet https://repo.anaconda.com/archive/Anaconda3-5.3.0-Linux-x86_64.sh -O ~/anaconda.sh && \
+    /bin/bash ~/anaconda.sh -b -p /opt/conda && \
+    rm ~/anaconda.sh && \
+    ln -s /opt/conda/etc/profile.d/conda.sh /etc/profile.d/conda.sh && \
```
dmorris-algo: For my understanding, are these three lines intended for interactive shell sessions where a person is exec'ing `/bin/bash` in the docker image?

As far as I can understand, these don't become available in `algorithmia-pipe` cause it doesn't run in a shell, which is why we have to repeat the [`loading and activation`](https://github.com/algorithmiaio/langpacks/pull/96/files#diff-445c51300ed7e5b6f9a8cae1e0a196a8R8). Correct me if I'm wrong, cause I'm trying to make sure I understand.

languages/anaconda3/template/requirements.yaml
dmorris-algo: Going to have to agree with Patrick here, we should probably not provide this as a customization option right now. By installing a fresh python version in each algorithm, we essentially invalidate any caching/sharing benefits we receive from Docker.

dmorris-algo (approved):
LGTM

libraries/anaconda35/Dockerfile
zeryx: Done, good call

libraries/anaconda35/Dockerfile
zeryx: You're right, I refactored these lines into the `algorithmia-build` script and verified functionality. This keeps the sourcing and shell ops away from docker which should make this process much less fragile.

languages/anaconda3/config.json
```diff
@@ -0,0 +1,7 @@
+{
+    "artifacts": [
+        {"source":"/home/algo/.conda", "destination":"/home/algo/.conda/"},
+	{"source":"/home/algo/.cache", "destination":"/home/algo/.cache/"},
```
jamesatha: nit spacing

languages/anaconda3/template/requirements.yaml
```diff
@@ -0,0 +1,13 @@
...
+- defaults
+dependencies:
+# Here are where your anaconda dependencies can be defined
+- python
+# If you wish to use a specific version of pip, please adjust the value below
+- pip=19.0
+- pip:
+# Add your pip dependencies (taken directly from pypi) here, we'll use the version of python and pip installed above
+  - six>=1.12.0
+  - algorithmia>=1.2,<1.3
```
jamesatha: Why are we restricting the upper end? I feel like this has bitten us in the past

libraries/anaconda35/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+    rm tini.deb && \
+    apt-get clean \
+    ; \
+    rm -rf /var/lib/apt/lists/*
+
+
+RUN wget --quiet https://repo.anaconda.com/archive/Anaconda3-5.3.0-Linux-x86_64.sh -O ~/anaconda.sh && \
+    /bin/bash ~/anaconda.sh -b -p /opt/conda && \
+    rm ~/anaconda.sh
+    #ln -s /opt/conda/etc/profile.d/conda.sh /etc/profile.d/conda.sh && \
```
jamesatha: Can we remove these comments? (and maybe the ones at the top of the file?)

libraries/anaconda35/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+RUN apt-get update --fix-missing && apt-get install -y wget bzip2 ca-certificates \
+    libglib2.0-0 libxext6 libsm6 libxrender1 \
+    curl grep sed dpkg \
+    git  && \
+    TINI_VERSION=`curl https://github.com/krallin/tini/releases/latest | grep -o "/v.*\"" | sed 's:^..\(.*\).$:\1:'` && \
+    curl -L "https://github.com/krallin/tini/releases/download/v${TINI_VERSION}/tini_${TINI_VERSION}.deb" > tini.deb && \
+    dpkg -i tini.deb && \
+    rm tini.deb && \
+    apt-get clean \
+    ; \
```
jamesatha: why a semi-colon? Can we use the `&&` on the previous line for consistency? (Or does this apt-clean fail? Also, don't you want the `-y` flag for clean?)

languages/anaconda3/template/requirements.yaml
zeryx: so that we can put potentially breaking changes in 1.3 that won't effect algorithms created in this new handshake - but I agree if we guarenteed that we followed semantic versioning this would be alright

libraries/anaconda35/Dockerfile
zeryx: I kept this consistent with what was done with java, would be down with pulling this out into a new issue

languages/anaconda3/template/src/Algorithm.py
```diff
@@ -0,0 +1,36 @@
...
+    return "hello {}".format(str(input))
+
+# Here is an example of an advanced form of an algorithm function,
+# which also uses a load function (see the example below)
+# -- ADVANCED ALGORITHM USAGE --
+# def apply(input, keras_model):
+#     prediction = keras_model.predict(input)
+#     result = {"class": prediction[0], "confidence": prediction[1]}
+#     return result
+
```
adnaanm: Very nit and not necessarily suggesting you bother changing it at this point, just for future reference, documentation can/should go in docstrigs

PR #97, opened 2019-07-22
Add git to python Dockerfile installs

adnaanm:
@pmcq this should solve the issue of not being able to use pip install to install from github because `git` isn't installed. Added JPeck for visibility

### DIFF
--- a/libraries/python36/Dockerfile
+++ b/libraries/python36/Dockerfile
@@ -15,7 +15,8 @@ RUN apt-get -y update && \
         libssl-dev \
         tk-dev \
         uuid-dev \
-        wget && \
+        wget \
+        git && \
     rm -rf /var/lib/apt/lists/*
 
 # MIT license from initial repo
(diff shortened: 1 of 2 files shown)


pmcq (commented):
have you tried building the image locally so we can be certain it works before merge? https://github.com/algorithmiaio/langpacks/blob/master/tools/packageset_validation.md

adnaanm:
> tried building the image locally so we can be certain it works before merge? https://github.com/algorithmiaio/langpacks/blob/master/tools/packageset_validation.md

Oh very cool, didn't know about this tool

adnaanm:
@pmcq built the image(s) using the tool and they're all good - would you be OK approving?

adnaanm:
@pmcq ping on this

pmcq (approved):

pmcq:
> @pmcq ping on this

my bad - usually am just looking at the queue of pull requests. if you click the `re-request review` button next to a name after doing updates it will show up in their queue again

adnaanm:
ah gotcha, thanks for the tip!

PR #100, opened 2019-08-06
DS-338 solaris dockerfile & template development

zeryx:
Machine learning template for GIS, installation process modelled after the guide defined here: https://github.com/CosmiQ/solaris
Built specifically for anaconda support, anaconda is required for this package to be compiled and bundled properly.


zeryx:
rebased to track https://github.com/algorithmiaio/langpacks/pull/103

templates/solaris-0.1.2/algorithmia.conf
```diff
@@ -0,0 +1,5 @@
+{
+    "username": "__USER__",
+    "algoname": "__ALGO__",
+    "language": "anaconda3x"
```
zeryx: should we change the language here? the package requires python 36, is `anaconda3x-python36` reasonable?

jamesatha: This should match a real language so what you have no seems good to me

templates/solaris-0.1.2/requirements.yml
```diff
@@ -0,0 +1,34 @@
+channels:
+  - pytorch
+  - conda-forge
+  - defaults
+dependencies:
```
jamesatha: Do we need this section? What happens when a user changes these?

templates/solaris-0.1.2/src/Algorithm.py
```diff
@@ -0,0 +1,37 @@
+import Algorithmia
+import solaris as sol
+# API calls will begin at the apply() method, with the request body passed as 'input'
+# For more details, see algorithmia.com/developers/algorithm-development/languages
+
+def apply(input):
+    return "hello {}".format(str(input))
```
jamesatha: I believe the `str` is unnecessary

libraries/anaconda3-solaris-0.1.2/context/solaris-build
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
```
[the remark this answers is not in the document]
zeryx: That works for me

templates/solaris-0.1.2/requirements.yml
zeryx: Just like with requirements.txt for pytorch, it allows users to know what's installed and have some feedback. If those variables change then a new build op will happen and the old versions may get evicted. A comment here similar to our other dependency based package sets might be a good idea ("Don't change this unless you need to")

templates/solaris-0.1.2/src/Algorithm.py
zeryx: it massages non-string types into something that can work, a copy/paste directly from the anaconda python template. I'm not really sure what a solaris template looks like, but I know this is a customer ask so I have a feeling they know what to do lol

libraries/anaconda3-solaris-0.1.2/context/solaris-build
```diff
@@ -0,0 +1,11 @@
...
+set -e
+
+. /opt/conda/etc/profile.d/conda.sh
+
+cd $1
```
jamesatha: A comment telling what $1 is and the fact that it should have an `environment.yml` file would be nice

libraries/anaconda3-solaris-0.1.2/context/solaris-build
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+set -e
```
jamesatha: Can we change this to `set -eu`? I feel like any script that expects the `ANACONDA_ENV` variable should error if that is not set

libraries/anaconda3-solaris-0.1.2/context/solaris-build
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+set -eu
```
jamesatha: I don't mind doing it in this PR or some other but it would be nice to make the same change in `libraries/anaconda3-buildtime/context/build` and `libraries/anaconda3-runtime/context/pipe` so they also barf when we have an unbound ANACONDA_ENV variable

PR #101, opened 2019-08-07
DS-338 provide support for anaconda modular dependencies

zeryx:
The original release of anaconda did not provide an endpoint to update the conda environment with new dependencies.

This PR allows for packages to update that conda environment without any substantial rework of the conda language packages.


libraries/anaconda3-buildtime/Dockerfile
```diff
@@ -1 +1,3 @@
 COPY --chown=0:0 anaconda3-buildtime/context/build /usr/local/bin/algorithmia-build
+RUN conda create --prefix /tmp/conda_environment && \
+    chmod --recursive ugo+rw /tmp/conda_environment
```
jamesatha: Should we change to `USER algo` before the `conda create` so permissions will be set for the algo users to read and write to that directory?

libraries/anaconda3-buildtime/context/build
```diff
@@ -2,9 +2,9 @@
 
 set -e
 
-CONDAENV=conda_environment
+CONDAENV=/tmp/conda_environment
```
jamesatha: why tmp? I feel like temporary storage should be temporary. Let's put this in a more permanent location. I'm cool with that being somewhere in `/home/algo`

jamesatha (commented):
Not a huge fan of the tmp dir use and don't know enough about conda to know if this is good practice

libraries/anaconda3-buildtime/Dockerfile
zeryx: use algo isn't defined at this point, that happens inside of build.j2 & compile.j2, would creating the user in advance in this instance mess anything up? I'm down with chowning to user algo:algo

libraries/anaconda3-buildtime/context/build
zeryx: works for me, /tmp is my default scratch space - but in this case /home/algo is fine (as long as the user is defined!?)

libraries/anaconda3-buildtime/Dockerfile
dmorris-algo: I presume you're referencing [Dockerfile.builder.j2](https://github.com/algorithmiaio/langpacks/blob/master/languages/Dockerfile.builder.j2), in which case I believe that `anaconda3-buildtime` would be templated in as a package [on line 17](https://github.com/algorithmiaio/langpacks/blob/master/languages/Dockerfile.builder.j2#L17), which is after the `algo` user is [defined above](https://github.com/algorithmiaio/langpacks/blob/master/languages/Dockerfile.builder.j2#L11). So, it should be defined at this point.

libraries/anaconda35/Dockerfile
```diff
@@ -21,7 +16,4 @@ RUN apt-get update --fix-missing && apt-get install -y wget bzip2 ca-certificate
 
 RUN wget --quiet https://repo.anaconda.com/archive/Anaconda3-5.3.0-Linux-x86_64.sh -O ~/anaconda.sh && \
     /bin/bash ~/anaconda.sh -b -p /opt/conda && \
-    rm ~/anaconda.sh
-    #ln -s /opt/conda/etc/profile.d/conda.sh /etc/profile.d/conda.sh && \
-    #echo ". /opt/conda/etc/profile.d/conda.sh" >> ~/.bashrc && \
-    #echo "conda activate base" >> ~/.bashrc
\ No newline at end of file
+    rm ~/anaconda.sh
```
dmorris-algo: Nit: add newline at EOF here/

dmorris-algo (commented):
Please re-request review after feedback has been incorporated. Thanks James!

libraries/anaconda3-buildtime/Dockerfile
```diff
@@ -1 +1,3 @@
 COPY --chown=0:0 anaconda3-buildtime/context/build /usr/local/bin/algorithmia-build
+RUN conda create --prefix ANACONDA_ENV && \
+    chown algo:algo -R ANACONDA_ENV
```
dmorris-algo: Nit: add newline at EOF

libraries/anaconda3-buildtime/Dockerfile
```diff
@@ -1 +1,3 @@
 COPY --chown=0:0 anaconda3-buildtime/context/build /usr/local/bin/algorithmia-build
+RUN conda create --prefix ANACONDA_ENV && \
```
jamesatha: So I'm confused here. `ANACONDA_ENV` isn't defined here, right?
Secondly, can we just become `USER algo` before this `RUN` operation and then we won't need to do the chown as part of the RUN command

libraries/anaconda3-buildtime/context/build
```diff
@@ -2,9 +2,9 @@
...
 set -e
 
-CONDAENV=conda_environment
+ANACONDA_ENV=/home/algo/conda_environment
 ROOTDIR=/opt/algorithm
 
 . /opt/conda/etc/profile.d/conda.sh
 
-conda env create -n $CONDAENV --file=$ROOTDIR/requirements.yaml -q --force
+conda env update --prefix ANACONDA_ENV --file=$ROOTDIR/requirements.yml
```
jamesatha: Did you mean `$ANACONDA_ENV`? Now, I believe you are using the string `ANACONDA_ENV`. I'm not sure how things are working - maybe `~/conda_environment` is the default location for the dependencies?
Why did we have the force before?

libraries/anaconda3-runtime/context/pipe
```diff
@@ -2,10 +2,10 @@
 
 set -e
 
-CONDAENV=conda_environment
+ANACONDA_ENV=/home/algo/conda_environment
 ROOTDIR=/opt/algorithm
 
 . /opt/conda/etc/profile.d/conda.sh
-conda activate $CONDAENV
+conda activate ANACONDA_ENV
```
jamesatha: Here too. You aren't interpreting the `ANACONDA_ENV` variable

libraries/anaconda3-buildtime/Dockerfile
dmorris-algo: Yeah, to follow up on that second point I've unresolved the initial thread: https://github.com/algorithmiaio/langpacks/pull/101#discussion_r312192350 and would appreciate clarification -- if only for my own understanding -- on why it would not be possible to use `USER algo`, if that is the case.

zeryx: environment variables set in dockerfiles propegate automatically into bash scripts,new commit will rectify this.

libraries/anaconda3-runtime/context/pipe
zeryx: fixed in latest version

dmorris-algo (approved):
LGTM

libraries/anaconda3-buildtime/Dockerfile
jamesatha: And the $ANACONDA_ENV should be a variable here as well. And wondering why we can't run this as user algo still

jamesatha: think this should be using the env variable instead of the literal

zeryx: user algo would disrupt this variable, I'd have to redeclare it after I switch to user algo. This is a variable, it's defined in the anaconda35 base image, if you guys would like I can redefine it at the top of both the runtime and the buildtime dockerfiles for legibility, but it would technically be redundant. Let me know what you think!

PR #103, opened 2019-08-12
Anaconda restructure

zeryx:
The version of anaconda that we released was not able to properly handle package dependencies prior to algorithm compilation. This PR includes:

- Implements a pattern for anaconda dependencies to get installed to the same environment and compile properly.
- Some refactors to reduce duplicate code and to improve legibility.
- Adds support for pre-installing specific versions of python (2.7, 3.5, and 3.7)


libraries/anaconda3-buildtime/context/build
```diff
@@ -4,6 +4,11 @@ set -e
...
-. /opt/conda/etc/profile.d/conda.sh
 
-conda env update --prefix $ANACONDA_ENV --file=$ROOTDIR/requirements.yml
+. /opt/conda/etc/profile.d/conda.sh
+if [ -z "$1" ]; then
+    conda env update --prefix "$ANACONDA_ENV" --file $ROOTDIR/requirements.yml;
+else
+    conda env update --prefix "$ANACONDA_ENV" --file "$1";
+    chown algo:algo -R $ANACONDA_ENV;
+fi
```
zeryx: Every anaconda dependency will use this build script internally, for example of use check out `anaconda3-python37/Dockerfile`
This keeps things DRY and allows us to reuse the same script rather than adding a ton of boilerplate.

libraries/anaconda3-python27/Dockerfile
```diff
@@ -0,0 +1,4 @@
+ENV REQFILE=/home/algo/dependency.yml
+COPY --chown=0:0 anaconda3-buildtime/context/build /usr/local/bin/dependency-build
+COPY --chown=0:0 anaconda3-python27/context/requirements.yml $REQFILE
+RUN dependency-build $REQFILE && rm $REQFILE
```
zeryx: This is an example of how one would install an anaconda dependency, the specific operation requested is described in the `requirements.yml` file - the rest is boilerplate and reuses the `build` file defined in `anaconda3-buidltime`

languages/anaconda3/template/requirements.yml
```diff
@@ -1,13 +1,11 @@
 channels:
 # Add your desired anaconda channels here
-- conda-forge
-- defaults
+  - conda-forge
+  - defaults
 dependencies:
 # Here are where your anaconda dependencies can be defined
-- python
 # If you wish to use a specific version of pip, please adjust the value below
```
jamesatha: Is this comment still valid?

libraries/anaconda3-buildtime/context/build
```diff
@@ -4,6 +4,11 @@ set -e
 
 ROOTDIR=/opt/algorithm
 
-. /opt/conda/etc/profile.d/conda.sh
 
-conda env update --prefix $ANACONDA_ENV --file=$ROOTDIR/requirements.yml
+. /opt/conda/etc/profile.d/conda.sh
+if [ -z "$1" ]; then
+    conda env update --prefix "$ANACONDA_ENV" --file $ROOTDIR/requirements.yml;
```
jamesatha: nit: why are there semi-colons at the end of these lines?
Also, I don't think you need the quotes around `$ANACONDA_ENV`

libraries/anaconda3-buildtime/context/build
```diff
@@ -4,6 +4,11 @@ set -e
 
 ROOTDIR=/opt/algorithm
```
jamesatha: Can we add a comment for what $1 is supposed to be

libraries/anaconda3-python27/Dockerfile
```diff
@@ -0,0 +1,4 @@
+ENV REQFILE=/home/algo/dependency.yml
```
jamesatha: I feel like this should be an `ARG` since it's not needed outside of builds

libraries/anaconda3-python27/Dockerfile
jamesatha: why do we remove the file? Not sure it is really necessary. I feel like it could help with debugging and it's such a small file that it won't matter from the space perspective

libraries/anaconda3-python35/Dockerfile
```diff
@@ -0,0 +1,4 @@
+ENV REQFILE=/home/algo/dependency.yml
```
jamesatha: I feel like these should be unique file locations instead of stomping over each other

languages/anaconda3/template/requirements.yml
zeryx: good catch I'll remove

libraries/anaconda3-buildtime/context/build
zeryx: I threw this though a shell linter and it recommended "'s to remove ambiguity, I'm down with removing em

libraries/anaconda3-buildtime/context/build
zeryx: yea that's not a bad idea, I'll add one

libraries/anaconda3-python35/Dockerfile
zeryx: yea that's why I was deleting the files at the end of the dockerfile ops, I was trying to keep things DRY but I think you're right, it's excessive. I'll add a suffix to the end like `dependency-python35.yml`

PR #106, opened 2019-08-16
fix shell scripts that fail with `set -eu` command

zeryx:
the changes requested by jamesA resulted in the following exception which were not returned by the package set construction process:
`/usr/local/bin/algorithmia-build: line 13: $1: unbound variable`
This seems to be due to the introduction of the `u` operator in the `set -eu` requested modification for build & solaris-build.

### DIFF
--- a/libraries/anaconda3-buildtime/context/build
+++ b/libraries/anaconda3-buildtime/context/build
@@ -1,6 +1,9 @@
 #!/bin/bash
 
-set -e
+
+FILE="$1"
+
+set -eu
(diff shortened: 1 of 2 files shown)


bytekast:
what is the `u` flag for?

jamesatha:
> what is the `u` flag for?

It's to make sure we crash if ANACONDA_ENV (or any other variable we expect to be set) isn't

jamesatha (approved):

PR #108, opened 2019-08-21
anaconda perf improvements (perf/disk size)

zeryx:
This PR replaces anaconda with miniconda, dramatically reducing disk utilization to improve buildtime/runtime performance of the anaconda based IPA images. We also removed the tini dependency which is not required for our use.

This reduces the anaconda base image by ~2.97GB, which further reduces later layer image sizes by being a necessary per-requesite layer.

### DIFF
--- a/libraries/anaconda35/Dockerfile
+++ b/libraries/anaconda35/Dockerfile
@@ -1,7 +1,6 @@
 ENV LANG=C.UTF-8 LC_ALL=C.UTF-8
 ENV PATH /opt/conda/bin:$PATH
 ENV ANACONDA_ENV=/home/algo/anaconda_environment
-ENV PYTHONDONTWRITEBYTECODE=true
 
 RUN apt-get update --fix-missing && apt-get install -y wget bzip2 ca-certificates \
     libglib2.0-0 libxext6 libsm6 libxrender1 \
--- a/libraries/anaconda35/Dockerfile
+++ b/libraries/anaconda35/Dockerfile
@@ -7,10 +7,6 @@ RUN apt-get update --fix-missing && apt-get install -y wget bzip2 ca-certificate
     libglib2.0-0 libxext6 libsm6 libxrender1 \
     curl grep sed dpkg \
     git  && \
-#    TINI_VERSION=`curl https://github.com/krallin/tini/releases/latest | grep -o "/v.*\"" | sed 's:^..\(.*\).$:\1:'` && \
-#    curl -L "https://github.com/krallin/tini/releases/download/v${TINI_VERSION}/tini_${TINI_VERSION}.deb" > tini.deb && \
-#    dpkg -i tini.deb && \
-#    rm tini.deb && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/*
 
--- a/libraries/anaconda35/Dockerfile
+++ b/libraries/anaconda35/Dockerfile
@@ -1,19 +1,19 @@
 ENV LANG=C.UTF-8 LC_ALL=C.UTF-8
 ENV PATH /opt/conda/bin:$PATH
 ENV ANACONDA_ENV=/home/algo/anaconda_environment
+ENV PYTHONDONTWRITEBYTECODE=true
 
 RUN apt-get update --fix-missing && apt-get install -y wget bzip2 ca-certificates \
     libglib2.0-0 libxext6 libsm6 libxrender1 \
     curl grep sed dpkg \
     git  && \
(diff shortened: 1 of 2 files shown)


zeryx:
We don't use the latest version of miniconda as it introduces an additional file that causes permissions issues that were not introduced with previous versions.

libraries/anaconda35/Dockerfile
```diff
@@ -1,19 +1,19 @@
...
     libglib2.0-0 libxext6 libsm6 libxrender1 \
     curl grep sed dpkg \
     git  && \
-    TINI_VERSION=`curl https://github.com/krallin/tini/releases/latest | grep -o "/v.*\"" | sed 's:^..\(.*\).$:\1:'` && \
-    curl -L "https://github.com/krallin/tini/releases/download/v${TINI_VERSION}/tini_${TINI_VERSION}.deb" > tini.deb && \
-    dpkg -i tini.deb && \
-    rm tini.deb && \
-    apt-get clean \
-    ; \
+#    TINI_VERSION=`curl https://github.com/krallin/tini/releases/latest | grep -o "/v.*\"" | sed 's:^..\(.*\).$:\1:'` && \
```
jamesatha: nit: I'd prefer removing the lines over commenting them out

libraries/anaconda35/Dockerfile
```diff
@@ -1,19 +1,19 @@
 ENV LANG=C.UTF-8 LC_ALL=C.UTF-8
 ENV PATH /opt/conda/bin:$PATH
 ENV ANACONDA_ENV=/home/algo/anaconda_environment
+ENV PYTHONDONTWRITEBYTECODE=true
```
jamesatha: Overall, I'm hoping this is fine but it does make me a bit nervous. I'm guessing load time might be a bit higher for these algorithms since imports will take a bit more time, but that is outweighed by the speed of downloading a smaller image?

libraries/anaconda35/Dockerfile
```diff
@@ -1,19 +1,19 @@
...
+#    dpkg -i tini.deb && \
+#    rm tini.deb && \
+    apt-get clean && \
     rm -rf /var/lib/apt/lists/*
 
-RUN wget --quiet https://repo.anaconda.com/archive/Anaconda3-5.3.0-Linux-x86_64.sh -O ~/anaconda.sh && \
-    /bin/bash ~/anaconda.sh -b -p /opt/conda && \
-    rm ~/anaconda.sh
+RUN wget --quiet https://repo.anaconda.com/miniconda/Miniconda3-4.6.14-Linux-x86_64.sh -O ~/miniconda.sh && \
+    /bin/bash ~/miniconda.sh -b -p /opt/conda && \
```
jamesatha: why do we have to start a new shell to run? Can we just `chmod +x ~/miniconda.sh` and run it?

templates/solaris-0.1.2/requirements.yml
```diff
@@ -4,30 +4,31 @@ channels:
...
+#  - rasterio=1.0.23
+#  - scipy=1.2.1
+#  - scikit-image=0.15.0
+#  - tensorflow=1.13.1
+#  - pytorch=1.1.0
+#  - torchvision=0.3.0
+#  - pyyaml=5.1
+#  - affine=2.2.2
+#  - albumentations>0.2.3
+#  - rio-cogeo=1.0.0
```
jamesatha: same here on the comments

libraries/anaconda35/Dockerfile
zeryx: honestly we can remove this. This was recommended from a "how to reduce anaconda dockerfile bloat"; but it was a very minor difference.

libraries/anaconda35/Dockerfile
zeryx: no, anaconda requires shells to work - being run in docker root causes all kinds of hell. This is taken directly from the miniconda dockerfile: https://hub.docker.com/r/continuumio/miniconda/dockerfile

templates/solaris-0.1.2/requirements.yml
zeryx: I commented all of this out to allow us to skip over "re-solving" the environment, which takes a considerable amount of time with anaconda.

pmcq: does it take a long time and then not do anything or is it also re-installing packages and stuff?

pmcq (approved):
:shrug:

templates/solaris-0.1.2/requirements.yml
zeryx: It would re-scan the local dependency tree for each dependency, and sometimes it would reinstall things that weren't necessary. For smaller package sets I'd probably leave that in there for posterity, but in this case every second counts and we want to reduce that compile time.

PR #110, opened 2019-09-03
DS-357  r-tensorflow-1.12 support

zeryx:
uses r as the base, with anaconda python for the tensorflow/keras/h5py interconnections with R.
Dockerfile and other components taken from here: https://github.com/tmobile/r-tensorflow-api/blob/master/Dockerfile


zeryx:
This project uses a version of anaconda that lives in global state, and is written to /etc/environment. Will migrate to other anaconda systems to keep consistency.

libraries/r36-buildtime/context/rip.py
```diff
@@ -55,15 +55,15 @@ def main():
...
                 sys.exit(1)
 
     if len(rscript) == 1:
-        print 'There is nothing to install'
+        print('There is nothing to install')
         return
 
     commandArgs = ['Rscript', '-e', '; '.join(rscript)]
-    print 'Running: {}'.format(' '.join(commandArgs))
+    print ('Running: {}'.format(' '.join(commandArgs)))
```
jamesatha: nit: remove space after `print`

libraries/r-tensorflow-1.12/Dockerfile
```diff
@@ -0,0 +1,15 @@
+# install miniconda, and set the appropriate path variables.
+RUN wget --quiet https://repo.anaconda.com/miniconda/Miniconda3-4.4.10-Linux-x86_64.sh -O ~/miniconda.sh && \
```
jamesatha: Can this use https://repo.anaconda.com/miniconda/Miniconda3-4.6.14-Linux-x86_64.sh so we can just use the anaconda35 library and not duplicate this?

zeryx: there is actually something different around the installation here, I might backport it into anaconda proper. It looks like I can install and run everything as root and get things to work in algorithms

zeryx: but that might only be because they are installed in the dockerfiles, I have a feeling they would run into issues when being installed by the user with algorithm-build

jamesatha (commented):
We talked in person about re-using the anaconda35 library so once that's done, we should be good to go

libraries/r-tensorflow-1.12/Dockerfile
```diff
@@ -0,0 +1,11 @@
+RUN echo ". /opt/conda/etc/profile.d/conda.sh" >> /etc/environment && \
+    echo "conda activate base" >> /etc/environment
+
+ENV PATH /opt/conda/bin:$PATH
+
+RUN conda install python=3.6 && \
+/opt/conda/bin/pip install tensorflow==1.12 && \
```
jamesatha: nit: i like what you did above by lining up the commands that are in the same RUN command

jamesatha (approved):

PR #111, opened 2019-09-20
Add pytorch-1.2.x dockerfile

besirkurtulmus:


besirkurtulmus:
> Would it be reasonable to also get torchvision installed as well? One of torchvisions pre-reqs is pytorch (for some reason) and can be a pain to install otherwise. Would be good to have a version of it pre-baked.

I tried compiling an algorithm by adding `torchvision` afterwards, and it successfully compile.

I also think it might be worth creating a separate packageset for that? Mostly because this has been the pattern we've been using for a while now for PyTorch packagesets.

PR #112, opened 2019-10-09
Solaris size reduction

zeryx:
This PR makes some changes that reduces the size of the solaris anaconda package set from ~6 GB to 3.81GB, this should hopefully be sufficient to improve incremental compilation enough to avoid the untagged docker image issues we're running into.


templates/solaris-0.1.2/src/Algorithm.py
```diff
@@ -3,16 +3,85 @@
 # API calls will begin at the apply() method, with the request body passed as 'input'
 # For more details, see algorithmia.com/developers/algorithm-development/languages
 
+import numpy as np
+import solaris as sol
+from shapely.wkt import loads
+
+def rescale_auto():
```
jamesatha: Maybe add some comments on what this does?

jamesatha (commented):
Works for me. Have we deployed it to test yet?

libraries/anaconda3-solaris-0.1.2/Dockerfile
```diff
@@ -1,3 +1,4 @@
-COPY --chown=0:0 anaconda3-solaris-0.1.2/context/solaris-build /usr/local/bin/solaris-build
-RUN git clone --single-branch --branch v0.1.2 https://github.com/CosmiQ/solaris.git /tmp/solaris && \
-    /usr/local/bin/solaris-build /tmp/solaris
+ENV REQFILE=/home/algo/dependency-solaris-0.1.2.yml
+COPY --chown=0:0 anaconda3-solaris-0.1.2/context/build /usr/local/bin/dependency-build
```
pmcq: can you name this something more specific to `solaris`? it feels weird to install a solaris-specific script to `/usr/local/bin` with a gneric name like this

zeryx: oh for sure, I didn't think the name would matter as it's just getting used once and thrown away, we could throw these in /tmp

pmcq: i don't see where the file would be getting thrown away, i think it would still be available in the output image (and since the file was `COPY`'d even deleting the file would only result in a tombstone marker)

zeryx: I moved to /tmp, it wouldn't be visible at the output image as these ops are done in the buildtime and aren't migrated. Let me know if that works

templates/solaris-0.1.2/src/Algorithm.py
zeryx: I have no idea what it does, this was provided to me as canary functions by the customer; this package set will never be public facing, so I don't see that as a terrible thing

libraries/anaconda3-solaris-0.1.2/Dockerfile
pmcq: i'm not sure i follow - ` it wouldn't be visible at the output image as these ops are done in the buildtime` the file would definitely be visible in the complete `langpack-builder` or `langpack-runner` image.  if the package is only installed in the buildtime image then it wouldn't be in the final algorithm image, but it's still a weird file to have sitting around that could be overwritten by any other package that writes data to `/tmp`.  i would just want the file named like `anaconda-solaris-build` or something, not `dependency-build`, regardless of the directory it's in, and putting in `/tmp` doesn't really make a difference since the file is left around anyway i think it's fine to be in `/usr/local/bin`, just need a unique filename

zeryx:
Can confirm this works on test as expected, dropped build time by 7 minutes. @pmcq @jamesatha any reservations from merging?

jamesatha (approved):
It's a little hard to have a sample that isn't well documented but otherwise it seems fine to me

PR #115, opened 2019-10-18
ALGO-670 anaconda renaming and update WIP

zeryx:
the anaconda package is actually versioned 4.x, the 3 flag actually denotes which version of python is default, 2 or 3.

We're updating the naming and also the version to the latest available, this PR will need corresponding changes in common, dockherder, and web.


tools/packageset_validator.py
```diff
@@ -49,7 +49,7 @@ def create_intermediate_image(docker_client, base_image, dependencies, workspace
     `Dockerfile.builder.j2` and `Dockerfile.runner.j2` in the languages directory.
     :param docker_client: The docker python client
     :param base_image: The base image type in which to stage your docker container from, defaults to "ubuntu:16.04", but can be any standard base image.
-    :param dependencies: A list of dependencies that this intermediate image depends on, excluding language components. (eg: pytorch-1.0.0, spacy-2.0.18, etc)
+    :param dependencies: A list of dependencies that this intermediate image depends on, excluding language components. (eg: python-torch-1.0.0, python-python-spacy-2.0.18, etc)
```
besirkurtulmus: There's duplicate python's here.

zeryx: oh good catch

PR #120, opened 2019-12-09
ALGO-596 tensorrt support

zeryx:
added and verified tensorRT support with onnx template.

My validator tool currently doesn't support gpus or algo keys, so please run the following command to build the package:
`./tools/packageset_validator.py -b nvidia/cuda:10.0-cudnn7-devel-ubuntu16.04 -g python3 -s python37 -d tensorrt-6.0-cuda10.0 -t dependency -n tensorrt-6.0-cuda10.0`
and then when that fails, run the final image with the following
`docker run -it --runtime=nvidia -p 9999:9999 -e ALGORITHMIA_API_KEY='YOUR_API_KEY_HERE' 773e57916fcb`, replacing `773e57916fcb` with your final image hash as an output from `docker images`.
**Requires NVIDIA Docker to be installed and configured to test**

### DIFF
--- a/libraries/tensorrt-6.0-cuda10.0/Dockerfile
+++ b/libraries/tensorrt-6.0-cuda10.0/Dockerfile
@@ -1,4 +1,4 @@
 # This is a unique packageset in which it requires both a specific version of CUDA, and a version of Ubuntu as well.
 COPY --chown=0:0 tensorrt-6.0-cuda10.0/context/install /tmp/install
-RUN wget -O - https://james-s-public.s3.amazonaws.com/TensorRT-6.0.1.5.Ubuntu-16.04.x86_64-gnu.cuda-10.0.cudnn7.6.tar.gz | (cd /tmp && tar zxf -)
+RUN wget -O - https://algorithmia-assets.s3.amazonaws.com/algo_dependencies/tensorrt/TensorRT-6.0.1.5.Ubuntu-16.04.x86_64-gnu.cuda-10.0.cudnn7.6.tar.gz | (cd /tmp && tar zxf -)
 RUN sh /tmp/install 6.0.1.5
--- a/libraries/tensorrt-6.0-cuda10.0/Dockerfile
+++ b/libraries/tensorrt-6.0-cuda10.0/Dockerfile
@@ -1,0 +1,4 @@
+# This is a unique packageset in which it requires both a specific version of CUDA, and a version of Ubuntu as well.
+COPY --chown=0:0 tensorrt-6.0-cuda10.0/context/install /tmp/install
+RUN wget -O - https://james-s-public.s3.amazonaws.com/TensorRT-6.0.1.5.Ubuntu-16.04.x86_64-gnu.cuda-10.0.cudnn7.6.tar.gz | (cd /tmp && tar zxf -)
+RUN sh /tmp/install 6.0.1.5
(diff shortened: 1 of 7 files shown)


libraries/tensorrt-6.0-cuda10.0/Dockerfile
```diff
@@ -0,0 +1,4 @@
+# This is a unique packageset in which it requires both a specific version of CUDA, and a version of Ubuntu as well.
+COPY --chown=0:0 tensorrt-6.0-cuda10.0/context/install /tmp/install
+RUN wget -O - https://james-s-public.s3.amazonaws.com/TensorRT-6.0.1.5.Ubuntu-16.04.x86_64-gnu.cuda-10.0.cudnn7.6.tar.gz | (cd /tmp && tar zxf -)
```
besirkurtulmus: I think we have a docker/wheels bucket specifically for pulling files. I would recommend putting it there instead of putting under `james-s-public`.

templates/tensorrt-6.0-cuda10.0/requirements.txt
```diff
@@ -0,0 +1,8 @@
+algorithmia>=1.0.0,<2.0
+six
+numpy>=1.16.0,<2.0
+pycuda==2018.1.1
+Pillow>=5.2.0
+wget>=3.2
+tensorrt==6.0.1.5 # unchangable
```
besirkurtulmus: Maybe might be worth expanding `unchangeable` a little bit more.

templates/tensorrt-6.0-cuda10.0/src/__ALGO__.py
```diff
@@ -0,0 +1,62 @@
+import Algorithmia
+import numpy as np
+import os
+from .auxillary import load, allocate_buffers, do_inference
+from PIL import Image
+# API calls will begin at the apply() method, with the request body passed as 'input'
+# For more details, see algorithmia.com/developers/algorithm-development/languages
+
+SIMD_ALGO = "util/SmartImageDownloader/0.2.14"
+MODEL_PATH = "data://algorithmiahq/collection/mobilenetv2-1.0.onnx"
```
besirkurtulmus: Is this file publicly accessible?

zeryx: it will be, good catch moving it there now

libraries/tensorrt-6.0-cuda10.0/Dockerfile
zeryx: What did you have in mind? Do we have any buckets that are accessable to us? (I no longer have our production aws account access)

besirkurtulmus: Yeah, we have a bucket called `algorithmia-docker` w/ a folder called `docker-deps` under it.

We should have read/write access assuming it hasn't been revoked – engineering has gone through a lot of changes over the last year, they may have accidentally revoked it.

besirkurtulmus: To be clear, the bucket is under the `prod` account.

templates/tensorrt-6.0-cuda10.0/src/__ALGO__.py
zeryx: its public now

PR #128, opened 2020-02-11
Add the apex-0.1 packageset

besirkurtulmus:


libraries/apex-0.1/Dockerfile
```diff
@@ -0,0 +1,7 @@
+ENV APEX_GIT_HASH 5b71d3695bf39efcdcda9dff5be2f70314b8f091
+RUN git clone https://github.com/NVIDIA/apex /tmp/apex
+WORKDIR "/tmp/apex"
+RUN git reset --hard $APEX_GIT_HASH
+RUN pip install -v --no-cache-dir --global-option="--cpp_ext" --global-option="--cuda_ext" ./
+RUN rm -rf /tmp/apex
```
kennydaniel: If we're going to delete tmp at the end, combine these steps with `&& \` so that we don't create intermediate layers

besirkurtulmus: Fixed.

PR #130, opened 2020-02-25
ALGO-XXX added firefox to installation settings, when users need ff support

zeryx:
Looks like we're missing a dependency for when users try and run the selenium package set with firefox.
`libdbus-glib-1.so.2: cannot open shared object file: No such file or directory`
From the docs, it looks like we'll need to add firefox as a hard dependency to overcome this issue.


besirkurtulmus:
> Looks like we're missing a dependency for when users try and run the selenium package set with firefox.
> `libdbus-glib-1.so.2: cannot open shared object file: No such file or directory`
> From the docs, it looks like we'll need to add firefox as a hard dependency to overcome this issue.

This was because we were only required to support Chrome, and decided not to support Firefox.

besirkurtulmus (approved):
LGTM.

platypii (approved):
Seems reasonable

PR #131, opened 2020-02-25
Algo-XXX add nvidia support to validator

zeryx:
This PR adds nvidia/cuda support to docker runtimes when requested by the user. Simply add the `--nvidia-support` flag, and ensure that you're base image extends/is `nvidia/cuda:xxx`.


zeryx:
rebased from master as it got out of sync.

PR #132, opened 2020-02-27
Use https for Maven Central resolver

adnaanm:
A 20.0 bugbash cluster (https://algo-200rc1.algo.click) is showing the following error when you try to build Scala algorithms:

```
Picked up JAVA_TOOL_OPTIONS: -Dfile.encoding=UTF8
Error during sbt execution: Invalid URL specified for 'my-ivy-proxy-releases': no protocol: None
```

I _think_ this may have to do with the Maven switching to https issue we had in January. I'm not sure if this is the problem, or if my changes here will solve it so I wanted to get some feedback from the team and and open to any testing strategies.

### DIFF
--- a/scala/template/build.sbt
+++ b/scala/template/build.sbt
@@ -15,8 +15,6 @@ mainClass in Compile := Some("algorithmia.Main")
 
 val repoUrl = System.getProperty("repo.url", "http://git.algorithmia.com")
 
-resolvers += "Maven Central" at "https://repo1.maven.org/maven2/org/"
-
 resolvers += "Typesafe Repository" at "http://repo.typesafe.com/typesafe/releases/"
 
 libraryDependencies ++= Seq(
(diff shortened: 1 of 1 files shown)


platypii (commented):
Actually I think we should just remove this line entirely. Maven central is included by default in SBT. Per:

https://www.scala-sbt.org/1.x/docs/Resolvers.html#Predefined+resolvers

jjbooth74 (dismissed):

jjbooth74 (approved):
Wanted to confirm that @platypii 's comment also applies to earlier SBT versions, because our [template uses 0.13](https://github.com/algorithmiaio/langpacks/blob/sbt-use-https/scala/template/bin/setup#L46).

Here's the docs confirming that maven is included in earlier versions as well:
https://www.scala-sbt.org/0.13/docs/Resolvers.html#Predefined+resolvers

PR #133, opened 2020-03-02
Algo-XXX automatic environment testing

zeryx:
This PR adds automatic environment testing to the packageset validator. It also adds an optional algorithmia API key environment variable that can be passed into the runtime container for algorithmia client API support.

I need feedback on:
Testing system (files, format, etc)
Structure and format (everything makes sense? sanity check)

Once I have that, I'll move forward on adding test support to all algorithm language templates & dependency templates.

TO TEST:
install new dependency pycurl (and any system dependencies required)
Run the following:
`./tools/environment_validator.py -g python3 -s python37 -t language -n python3 -a 1`

This should return a success as it validates two test cases

PR #134, opened 2020-03-03
Fix the apex installation

besirkurtulmus:
This fixes the apex packageset/environment.

This is the command to run the environment: `./tools/packageset_validator.py -g python3 -s python37 -d pytorch-1.4.x -d apex-0.1 -t dependency -n apex-0.1 -b nvidia/cuda:10.1-cudnn7-runtime-ubuntu16.04 --nvidia-support 1`

This is the command to test the algorithm endpoint: `curl localhost:9999 -H 'Content-Type: application/json' -d '{"vector": [[0, 1, 2, 3],[4, 5, 6, 7],[8, 9, 10, 11],[12, 13, 14, 15]]}'`

### DIFF
--- a/templates/apex-0.1/src/__ALGO__.py
+++ b/templates/apex-0.1/src/__ALGO__.py
@@ -1,61 +1,37 @@
 import Algorithmia
 import torch as th
-import apex
+from apex import amp
 
 """
 This package set comes preinstalled with Nvidia Apex 0.1
 
 Example Input:
 {
-    "matrix_a": [[0, 1], [1, 0]],
-    "matrix_b": [[25, 25], [11, 11]]
-}
-
-Expected Output:
-{
-    "product": [[11, 11], [25, 25]]
+    "vector": [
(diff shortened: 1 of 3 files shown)


templates/apex-0.1/src/__ALGO__.py
```diff
@@ -1,9 +1,61 @@
...
+    "matrix_b": [[25, 25], [11, 11]]
+}
+
+Expected Output:
+{
+    "product": [[11, 11], [25, 25]]
+}
 """
 
+device = th.device('cuda:0' if th.cuda.is_available() else 'cpu')
```
zeryx: if we're gonna test gpu support, we shouldn't default to CPU - we should make sure that GPU support works, otherwise there's not much reason to do testing

besirkurtulmus:
@zeryx I merged in your https://github.com/algorithmiaio/langpacks/pull/133 PR into this one, and added a working template.

PR #139, opened 2020-03-30
ALGO-756 github actions CI template

zeryx:
Only useful when backed by github, automatically links to github actions.

TO TEST:
- Create a regular python37 algorithm on test, copy the `.github` directory and `TEST_CASES.json` into the new algorithm repo.
- Create or get an API Key that is able to call your algorithms + manage them
- Go to the github repo, click settings -> secrets and add a new secret called "api_key", paste the key from step 2 here
- if testing via test, edit .github/workflows/main.yml, replacing the variable for `api_address` with `https://api.test.algorithmia.com` instead of `https://api.algorithmia.com`
- commit your changes and check the actions tab for success/fail


platypii:
There is a part of me that wonders if this really belongs as a template, since it's the sort of thing that gets applied to any template really, not a template on it's own? I'm okay with it for internal testing purposes, but something about merging this into langpacks feels weird.

zeryx:
> There is a part of me that wonders if this really belongs as a template, since it's the sort of thing that gets applied to any template really, not a template on it's own? I'm okay with it for internal testing purposes, but something about merging this into langpacks feels weird.

I'd like to go and add this to all existing templates. If we automatically add this during the github repo creation I think we should still have the code live somewhere. What do you think?

PR #140, opened 2020-04-16
Attempt to fixup the last of the compile errors via guess and check

pmcq:
This seems to get the information plumbed through, the error message still shows up as `exited with code 1` but the stderr/stdout get populated and we've previously had code that will scan for that one error message and instead report back stderr.

My thought is that it could make sense to just tack the stderr/stdout directly to the `error message` during unexpected exit but this at least solves the immediate problem (with a few modifications in other components)

### DIFF
--- a/src/langrunner.rs
+++ b/src/langrunner.rs
@@ -226,7 +226,7 @@ impl LangRunnerProcess {
                 match line_result {
                     Ok(line) => {
                         match arc_stderr_buf.lock() {
-                            Ok(mut lines) => lines.push_str(&line),
+                            Ok(mut lines) => lines.push_str(&line), // TODO: Add newline character
 						    Err(err) => error!("{} {} Failed to get lock on stderr buffer: {}", LOG_IDENTIFIER, req_id, err),
                         }
                         info!("{} {} {}", "ALGOERR", req_id, line);
@@ -258,7 +258,7 @@ impl LangRunnerProcess {
                         break;
                     } else {
                         match arc_stdout_buf.lock() {
-                            Ok(mut lines) => lines.push_str(&line),
+                            Ok(mut lines) => lines.push_str(&line), // TODO: Add newline character
 						    Err(err) => error!("{} {} Failed to get lock on stdout buffer: {}", LOG_IDENTIFIER, req_id, err),
                         }
                         info!("{} {} {}", "ALGOOUT", req_id, &line);
--- a/src/langrunner.rs
+++ b/src/langrunner.rs
@@ -149,7 +149,8 @@ impl LangRunner {
             };
 
             // Augment output with duration and stdout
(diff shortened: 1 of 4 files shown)


src/main.rs
```diff
@@ -55,22 +55,28 @@ fn main() {
...
     match listener_res {
         Ok(mut listener) => {
-            let _ = load_complete(Ok(()), duration).or_else(|_| listener.close());
+            let _ = load_complete(Ok(()), duration, None, None).or_else(|_| listener.close());
         }
         Err(err) => {
             error!("{} {} Failed to load: {}", LOG_IDENTIFIER, "-", err);
-            let _ = load_complete(Err(err), duration);
+            let _ = match err {
+                Error::UnexpectedExit(code, stdout, stderr) => load_complete(
```
anowell, line 63:
```
match err {
    err @ Error::UnexpectedExit(code, stdout, stderr) => load_complete(Err(err.clone()), ...etc...),
    err => load_complete(Err(err), ...etc...),
}
```

but it's really not much worse than what you did an reiterates why the error handling in langserver is poorly designed

pmcq: i think i had tried `err.clone()` and got this:
```
err.clone()
   |                         ^^^^^ method not found in `langserver::error::Error`
```
so i just re-created the object :shrug:.  I'm doing a little more digging into just adding stderr directly to the `UnexpectedExit`'s `message` since i think that would end up waaaay simpler and I think accomplish the same goal

src/langrunner.rs
```diff
@@ -226,7 +226,7 @@ impl LangRunnerProcess {
                 match line_result {
                     Ok(line) => {
                         match arc_stderr_buf.lock() {
-                            Ok(mut lines) => lines.push_str(&line),
+                            Ok(mut lines) => lines.push_str(&line), // TODO: Add newline character
```
anowell, line 229:
```
{
lines.push_str(&line);
lines.push('\n');
}
```

pmcq: yup thanks, confirmed that fixed it

pmcq:
Confirmed that with this build of langserver I was able to get error message & stacktrace piped back through as expected.

src/main.rs
pmcq: checked into this again and Error doesn't have the #[derive CLone] and moreover can't have it because some of the subtypes don't have clone either:
```
error[E0277]: the trait bound `serde_json::Error: std::clone::Clone` is not satisfied                                                           
error[E0277]: the trait bound `hyper::client::Response: std::clone::Clone` is not satisfied
                                                                                                                                                      error[E0277]: the trait bound `std::io::Error: std::clone::Clone` is not satisfied                                                         
error[E0277]: the trait bound `hyper::Error: std::clone::Clone` is not satisfied                                                                      
```

PR #141, opened 2020-04-17
Update langserver to query stdout/stderr for algorithms

pmcq:
Prior to this, langserver would always send an empty `stdout` and `stderr` in the event of unexpected error conditions. Notably, this includes when an algorithm crashes during a load step in which case processes will exit (i.e. will not write anything to the FIFO) and we need to capture stderr in order to return back useful error messages.  

See exception handling here:
https://github.com/algorithmiaio/langpacks/blob/master/java/pipe/src/main/java/com/algorithmia/Pipe.java#L50-L60
https://github.com/algorithmiaio/langpacks/blob/master/python/template/bin/pipe.py#L14

(I know these are the old templates but the new ones are identical)


src/langrunner.rs
```diff
@@ -205,28 +216,40 @@ impl LangRunnerProcess {
...
                     .clone().unwrap_or_else(|| "-".to_owned());
                 match line_result {
-                    Ok(line) => info!("{} {} {}", "ALGOERR", req_id, line),
+                    Ok(line) => {
+                        match arc_stderr_buf.lock() {
+                            Ok(mut lines) => {
+                                lines.push_str(&line);
+                                lines.push('\n');
+                            }
+						    Err(err) => error!("{} {} Failed to get lock on stderr buffer: {}", LOG_IDENTIFIER, req_id, err),
```
jamesatha, line 233: nit: indentaton looks off cause of tabs

src/langrunner.rs
```diff
@@ -205,28 +216,40 @@ impl LangRunnerProcess {
...
+                        }
+                        info!("{} {} {}", "ALGOERR", req_id, line);
+                    },
                     Err(err) => error!("{} {} Failed to read line: {}", LOG_IDENTIFIER, req_id, err),
                 }
             }
         });
 
         let mut reader = BufReader::new(stdout);
+        let arc_stderr_buf = stderr_buf.clone();
```
jamesatha, line 243: Am I reading this right or do you have 2 variables with the same name in this scope?

src/notifier.rs
```diff
@@ -19,7 +18,7 @@ impl Notifier {
     pub fn parse(url: &str) -> Result<Notifier, Error> {
         match Url::parse(url) {
             Ok(parsed_url) => Ok(Notifier { url: parsed_url }),
-            Err(err) => Err(Error::UrlParseError(err.description().to_owned())),
+            Err(err) => Err(Error::UrlParseError(err.to_string())),
```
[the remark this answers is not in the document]
anowell: `description()` was [deprecated](https://doc.rust-lang.org/std/error/trait.Error.html#method.description).

`Error` types already have to implement `Display` which makes them printable and provides a `.to_string()` method. That method is more flexible for formatting error messages than `description` (returning `String` vs `&str`). In practice, `description()` was almost always used to return a hard-coded static string representation of the type (e.g. "URL Parse Error") and it's usage almost always indicated someone mistakenly thought `description()` was the right way to print an error message (like the very commit where I used `description()` here 4 years ago). So it was deprecated and now shows up as a compile warning.

src/langrunner.rs
anowell: suggest running `cargo fmt` on in a separate commit or PR - designed to keep formatting consistent without thinking about it, but it will probably touch a lot of lines the first time it's run.

src/langrunner.rs
anowell: yep. they're also both references to the same data, as both are `Arc`s cloned from the same underlying variable. Functionally, you can't use the first `arc_stderr_buf` after it was "moved" into a thread, so this just makes an extra `arc_stderr_buf` here.

It might seem strange, but it's not uncommon for rust code to create a new `let-binding` to a previously used variable name that won't be used again; the compiler catches the footguns that make this scary in most languages. But it is equally reasonable to just name them `arc_stderr_buf_1` and `arc_stderr_buf_2`.

src/langrunner.rs
pmcq: we can run `cargo fmt` separately to help clarify these issues, not sure how that ended up happening here.

anowell: that was me editting in `vim` without a vimprofile to use spaces instead of tabs... :-(

PR #142, opened 2020-04-20
Migrate to gitlab CI

kennydaniel:


ZachEddy:
I think we can condense this into just a `test` stage to reduce CI time. The `build` stage just downloads dependencies and caches them for the `test` stage, so the context switching between stages isn't super necessary and adds time.

Looking at the docs [here](https://doc.rust-lang.org/book/ch14-01-release-profiles.html#customizing-builds-with-release-profiles), the `--release` flag tells `cargo` to optimize compilation (I assume that makes the resulting binary smaller). IMO this would make sense if we published an artifact with the binary, but it looks like we discard the binary once the build finishes.

Since we don't really need the `--release` flag, I think we can just run `cargo test` in one stage because it will compile the code before testing it.

kennydaniel:
8m31s down to 1m33s good call @ZachEddy

PR #144, opened 2020-04-28
Fixup stdio reporting when the algorithm ends up exiting the process 

pmcq:
since we read stdio multiple times which will result in empty strings being reported back

Confirmed locally that this will now do the correct thing in the event of a system.exit(1) inside of algorithm execution.  it's much more difficult to test the async load failure message locally so will validate that in cluster tomorrow prior to final merge.  The error stuff is really awkward and I'm not sure there's a better approach since Error is not clone-able (see https://github.com/algorithmiaio/langpacks/pull/140/files#r416299280).  Also the fact that stdio is only read for UnexpectedExit and nothing else contributes to this being awkward among other things but... whatever.

### DIFF
--- a/src/message.rs
+++ b/src/message.rs
@@ -79,9 +79,18 @@ impl RunnerState {
                 }
             }
             RunnerState::Exited(err) => {
+                // If the err type was an UnexpectedExit then we would already have stdout and
+                // stderr from the error message so default to that instead
+                // The Error enum doesn't support .clone() so have to do some weird object
+                // re-creations here
+                let err_stdio = match err {
+                    Error::UnexpectedExit(code, stdout, stderr) =>
+                        (Error::UnexpectedExit(code, stdout.clone(), stderr.clone()), Some(stdout.clone()), Some(stderr.clone())),
+                    _ => (err, stdout, stderr)
+                };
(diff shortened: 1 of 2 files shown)


pmcq:
Okay, so I tested this out in cluster and discovered that it doesn't seem to fix the issue in the async work case, however this is actually an issue in our old system as well as the new one so I'm going to punt on trying to fix this any further.  Will document findings

src/message.rs
```diff
@@ -79,9 +79,18 @@ impl RunnerState {
                 }
             }
             RunnerState::Exited(err) => {
+                // If the err type was an UnexpectedExit then we would already have stdout and
+                // stderr from the error message so default to that instead
+                // The Error enum doesn't support .clone() so have to do some weird object
```
anowell, line 84: Not sure if you can add `#[derive(Clone)]` annotation to the current Error enum. I think it might of the random errors it wraps might be uncloneable, making it something that we can't `#[derive(Clone)]` for - (part of why I wanted to see the entire error type replaced with something simpler... anyhow, no concrete suggestion here, more of the thing that might drive me crazy enough to tinker with a refactor on top of your PR. :-D

PR #148, opened 2020-06-01
[ALGOENV-187] Use custom CA certificates when building IPA algorithms

dmorris-algo:
Relates to: https://github.com/algorithmiaio/containers/pull/171

- All references to UID 2222 updated to user `algo` in `Dockerfile.compile.j2` for consistency
  - `customize-container.sh` requires a username to `su` to
  - We only officially support the `docker` build mode; the `kaniko` build mode is not officially supported. Kaniko was the motivation behind using UIDs instead of usernames.
- Update should be transparent to @zeryx's local development/testing of new IPA environments

### DIFF
--- a/languages/Dockerfile.compile.j2
+++ b/languages/Dockerfile.compile.j2
@@ -1,19 +1,27 @@
 FROM {{builder_image}} as builder
-USER 2222
+
 # Docker build commands don't resolve environment variables so need this to either be numeric or a build argument
-COPY --chown=2222:2222 algosource /opt/algorithm/
+COPY --chown=algo:algo algosource /opt/algorithm/
 
 {% if config.local_dependency_src_path is defined %}
-COPY --chown=2222:2222 {{config.local_dependency_src_path}} {{config.local_dependency_dest_path}}
+COPY --chown=algo:algo {{config.local_dependency_src_path}} {{config.local_dependency_dest_path}}
 {% endif %}
 
 ENV HOME=/home/algo
+{% if config.local_dependency_src_path is defined %}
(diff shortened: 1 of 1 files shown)


languages/Dockerfile.compile.j2
```diff
@@ -1,19 +1,27 @@
...
-COPY --chown=2222:2222 algosource /opt/algorithm/
+COPY --chown=algo:algo algosource /opt/algorithm/
 
 {% if config.local_dependency_src_path is defined %}
-COPY --chown=2222:2222 {{config.local_dependency_src_path}} {{config.local_dependency_dest_path}}
+COPY --chown=algo:algo {{config.local_dependency_src_path}} {{config.local_dependency_dest_path}}
 {% endif %}
 
 ENV HOME=/home/algo
+{% if config.local_dependency_src_path is defined %}
```
dmorris-algo, line 11: Using the `local_dependency_src_path` property for build environment detection since @zeryx uses this for his local IPA dev/test workflow. In the case where we're building locally, we won't have `customize-container.sh` available, nor will any custom CA certs be available.

If there's a better way to detect or improve this flow, we can update in the future.

pmcq (approved):
looks good to me

languages/Dockerfile.compile.j2
zeryx: This is only necessary when building new language packages so that you don't have to publish all of your intermediate changes to a central repository. Honestly going forwards I think we could remove this, however everything LGTM

zeryx (approved):

PR #157, opened 2020-08-21
template file cleanup

zeryx:
turns out when I added the automatic language testing to the environment validator script, I put the test files _inside_ of the algorithm templates, which means they were exported to actual real algorithm environments. Not what was intended, this change fixes that and ensures that the environment test files are not bundled into new algorithm projects.


aslisabanci (approved):
well at least you noticed @zeryx :)

PR #158, opened 2020-08-24
[PSE-245] Package mirror arguments for use in IPA build and execution

dmorris-algo:
Five package mirror arguments are added to the _builder_ image; `CRAN_MIRROR` is passed into `langserver`'s environment during algorithm execution.

This supports both IPA algorithm builds and IPA algorithm executions:
- https://github.com/algorithmiaio/containers/pull/185
- https://github.com/algorithmiaio/docker-builder/pull/1
- https://github.com/algorithmiaio/node-autoscaler/pull/148

### DIFF
--- a/languages/Dockerfile.compile.j2
+++ b/languages/Dockerfile.compile.j2
@@ -1,4 +1,9 @@
 FROM {{builder_image}} as builder
+ARG CRAN_MIRROR
+ARG IVY_MIRROR
+ARG MAVEN_MIRROR
+ARG NPM_REGISTRY
+ARG PYTHON_INDEX_URL
 
 # Docker build commands don't resolve environment variables so need this to either be numeric or a build argument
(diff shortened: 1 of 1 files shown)


pmcq:
don't worry about dockherder.  but also in general - why do we need the mirror args at runtime? I would expect that users aren't installing more packages at runtime than they already did at compile.  i think this might've been there because of some weirdness with R specifically, but the other languages can probably be removed (and hopefully we can remove R's too since it's sloooow)

zeryx (dismissed):
LGTM, checked all workflows with `./tools/environment_validator.py` and nothing's broken in my world.

aslisabanci:
Tested the branch by running a previously successful "environment validator" script run to build a Pytorch package and it was successful.

robert-close (dismissed):

pmcq (approved):
:taco: :taco: :taco:

PR #159, opened 2020-08-26
[PSE] Add (Environments) designation to language display names

dmorris-algo:
The Algorithm Environment Manager gets the list of languages and their configuration from Legit on startup, which uses the language `config.json` files from this repository. In this PR, I'm updating the display names to match the manual fixup that was done during our release to test and production marketplace. The "(Environments)" designation will disappear over the next few sprints, once we settle upon a migration path from legacy to the new algorithm environments, with input from the product team.

Screenshot from 2020-08-26 16-38-16

### DIFF
--- a/languages/anaconda3/config.json
+++ b/languages/anaconda3/config.json
@@ -1,5 +1,5 @@
 {
-    "display_name": "Anaconda 3.5x",
+    "display_name": "Anaconda 3.5x (Environments)",
     "artifacts": [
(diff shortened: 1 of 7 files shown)


zeryx (approved):
LTGM, I thought this was taken from the registration files in envpacks, good to know!

dmorris-algo:
> LTGM, I thought this was taken from the registration files in envpacks, good to know!

Nope, not right now, and we expect to clear this ambiguity up over the next few sprints as well. :smile:

PR #166, opened 2020-10-16
Added tensorflow-cpu-2.3 to libraries and templates

aslisabanci:
Hey James, as per our discussion today, I added `tensorflow-cpu-2.3` to the `libraries` and the `templates` directories. 
The template code doesn't include the GPU allocation snippets and and the library Dockerfile is set to install `tensorflow` between v2.3 and 2.4

While I was on this, I realized one thing about our `tensorflow-gpu-2.3` package and I'll put a comment on the related line now. Can you please double check that line too? If you think that should be changed, I'll send another commit for it before we merge this branch. 

Thanks as always! :)


aslisabanci:
@zeryx What I was talking about is this line: https://github.com/algorithmiaio/langpacks/blob/9a5201cc421282c6149a3a2a4482e48b0335c590/libraries/tensorflow-gpu-2.3/Dockerfile#L2

Do you think we should've said `RUN pip install "tensorflow-gpu>=2.3,<2.4"` instead?

templates/tensorflow-cpu-2.3/algorithmia.conf
```diff
@@ -0,0 +1,5 @@
+{
+    "username": "__USER__",
+    "algoname": "__ALGO__",
+    "language": "python-3.x-1"
```
zeryx, line 4: I believe going forwards this should be `python3` right @pmcq ?

zeryx (commented):
Everything looks great, I like that you added the test to the actual test page (we were trying to figure out whether or not we wanted to use that test file with our templates.. but if it's there we should use it), the only thing I want clarification on is the language we define above, and I'll need @pmcq  or @kennydaniel to weigh in on that one.

zeryx (approved):
LGTM

PR #172, opened 2020-11-17
Caching of dependencies 

danielfrg:
Main change of this PR is the the new `req_files`  option in the `config.json` file of the languages, (see python for example).

This allows us to install `COPY` these files and run `algorithmia-build` before `COPY`ing the entire algo source,
that way changes on files in the algo source that are not listed in `req_files` will not trigger an installation of the dependencies.

Other changes in the validator code:
- Fixes some jinja templates whitespace issues that were causing warnings (and future failures) on `docker build`
- Added the ability to see the progress of the `docker build` commands with the `--verbose/-v` flag
- Make the cleanup of the generated dockerfiles explicit - makes it a bit easier to debug
- Removed the uuid on the docker image name to make it explicit based on the `dependencies` - not sure if this was intentionally for a reason but it was a bit annoying to see 100s of docker images :D 
- There are also formatting changes that happened automatically based on my editor config but we can revert those if needed


zeryx (changes requested):
Works fine for python, but fails with other languages that require compilation:

for a java11 config.json:
```
{
    "display_name": "Java OpenJDK 11.0 (Environments)",
    "req_files": [
        "pom.xml"
    ],
    "artifacts": [
        {"source":"/opt/algorithm/target/*.jar", "destination":"/opt/algorithm/target/algorithm.jar"},
        {"source":"/opt/algorithm/target/lib", "destination":"/opt/algorithm/target/lib/"}
    ]
}


built with:
`./tools/environment_validator.py -s java11 -t language -n java11 -b ubuntu:20.04`

```
build process fails with the following message: 
```
Error: Could not find or load main class com.algorithmia.algorithm.Algorithm\n2020-11-23T18:13:03.104 INFO ALGOERR - Caused by: java.lang.ClassNotFoundException: com.algorithmia.algorithm.Algorithm\n'
```

^ the if/else switch in the jinja template does work; ideally if possible a more generalized solution would be worth implementing here.

danielfrg:
(On the last commit) I was able to get it working with:

```
{
    "display_name": "Java OpenJDK 11.0 (Environments)",
    "req_files": [
        "pom.xml",
        "src"
    ],
    "artifacts": [
        {"source":"/opt/algorithm/target/*.jar", "destination":"/opt/algorithm/target/algorithm.jar"},
        {"source":"/opt/algorithm/target/lib", "destination":"/opt/algorithm/target/lib/"}
    ]
}
```

On this case it wont do anything tho, as any changes in the algo source will trigger a rebuild of that layer.
I don't think i see a way to make this work for java since we are calling `mvn package` it will require the source.

We could change the build command to just `mvn install dependency:copy-dependencies` and move the `mvn package` to the runtime but it wouldn't really be runtime at that point.

zeryx (dismissed):
Fair enough @danielfrg , I think what we have now is sufficient to move forward with. Lets just add a `req_file` for python2, python3, and anaconda3 and we should be good to merge this.

zeryx (approved):

PR #173, opened 2020-11-20
Add anaconda-49 language

danielfrg:
Taking just the conda langpack changes from https://github.com/algorithmiaio/langpacks/pull/172

Adding a new language pack for miniforge3.

Miniforge is the Miniconda alternative from conda-forge
- Using this instead in order to be ok with the new license from Anaconda: https://www.anaconda.com/blog/anaconda-commercial-edition-faq 
- This pack uses mamba as a package manager instead of conda  - mamba is compatible with conda and a lot faster and its already being used in systems like conda-forge and binder
- Added a package conda-solaris with the dependencies of solaris to test the new language uses environment.yml as expected

Test:

```
./tools/environment_validator.py -g conda-forge3 -s conda-forge3-4.9.0 -t dependency -n conda-solaris
```

cc @zeryx


zeryx (commented):
Verified this does work, will approve after a rename to `anaconda49`

danielfrg:
Renamed to `anaconda49`

```
./tools/environment_validator.py -g anaconda3 -s anaconda49 -t dependency -n anaconda-solaris
```

languages/anaconda3/template/algorithmia.conf
```diff
@@ -1,5 +1,5 @@
 {
     "username": "__USER__",
     "algoname": "__ALGO__",
-    "language": "anaconda3x"
+    "language": "conda3x"
```
zeryx: hmm this is a difficult one, we should call it 4x, but changing this would require also modifying additional locations in `algorithmiahq/common`, etc which would require multiple redeploys... Lets keep this as `anaconda3x` for now, and continue testing in test + caching logic built out in the sister PR. If everything works then I would argue for changing the language and modifying the upstream systems.

zeryx (approved):

languages/anaconda3/template/algorithmia.conf
danielfrg: Agree, done.

PR #178, opened 2020-12-14
ALGO-890 added lzma to all python base images

zeryx:
URGENT:
lzma support is a blocker for one of our customers.
# To Test
` ./tools/environment_validator.py -g python[x] -s python[x] -t language -n python[x]`
^ please test the construction of each version of python that we support:
* python27
* python36
* python37
* python38

Please also verify with a simple `import pandas as pd` that we no longer get the lzma warning, to sanity check that the `lzma` package is installed.

### DIFF
--- a/libraries/anaconda35/Dockerfile
+++ b/libraries/anaconda35/Dockerfile
@@ -3,7 +3,7 @@ ENV PATH /opt/conda/bin:$PATH
 ENV ANACONDA_ENV=/home/algo/anaconda_environment
 
 RUN apt-get update --fix-missing && apt-get install -y wget bzip2 ca-certificates \
-    libglib2.0-0 libxext6 libsm6 libxrender1 \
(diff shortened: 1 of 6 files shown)


aslisabanci:
✅Tested creating the images with:
```
./tools/environment_validator.py -g python2 -s python27 -t language -n python2

./tools/environment_validator.py -g python3 -s python36 -t language -n python3

./tools/environment_validator.py -g python3 -s python37 -t language -n python3

./tools/environment_validator.py -g python3 -s python38 -t language -n python3

```

✅ After the final images are created, tested the running container with:
`curl localhost:9999 -H 'Content-Type: application/json' -d '{"name": "Anthony"}'`

✅  docker exec'd into the final containers to do: 
- `pip install pandas`
- `import pandas as pd`
- observed no warnings from the python executor

aslisabanci (approved):

PR #179, opened 2020-12-21
Add pipe script for anaconda3

danielfrg:
I saw the file was missing while testing the new language on the test environment.

I think this type of errors are hidden from the validator script because of how we handle the docker context here.

Test with:

```
./tools/environment_validator.py -b ubuntu:18.04 -n anaconda3 -g anaconda3 -s anaconda49 -t language
```

### DIFF
--- a/languages/anaconda3/config.json
+++ b/languages/anaconda3/config.json
@@ -1,5 +1,5 @@
 {
-    "display_name": "Python (Conda) Environments",
+    "display_name": "Conda Environments - beta",
     "req_files": [
         "environment.yml"
     ],
--- a/languages/anaconda3/config.json
+++ b/languages/anaconda3/config.json
@@ -1,5 +1,5 @@
 {
-    "display_name": "Conda Environments - beta",
+    "display_name": "Conda (Environments) - beta",
     "req_files": [
         "environment.yml"
     ],
--- a/languages/anaconda3/template/src/Algorithm.py
+++ b/languages/anaconda3/template/src/Algorithm.py
@@ -1,0 +1,36 @@
+import Algorithmia
+# API calls will begin at the apply() method, with the request body passed as 'input'
+# For more details, see algorithmia.com/developers/algorithm-development/languages
+
+def apply(input):
+    return "hello {}".format(str(input))
+
+# Here is an example of an advanced form of an algorithm function,
+# which also uses a load function (see the example below)
+# -- ADVANCED ALGORITHM USAGE --
+# def apply(input, keras_model):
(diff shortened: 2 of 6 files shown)


zeryx:
Good find, so that's an interesting piece as we have the "New Developer Experience" system that exists here( https://github.com/algorithmiaio/algorithmia-python/blob/master/Algorithmia/handler.py) that Anaconda environments use, but vanilla pypi python does not - however that's not due to a technical limitation but rather product weigh-in. The `pipe` operation is a little different here, but https://github.com/algorithmiaio/langpacks/blob/2b0ac88804aaca2fae04d07f2f9f1149225b4ba4/libraries/anaconda3-runtime/context/pipe defines what that script would look like; as you can see it's simply excuting the algorithm primary file as an executable.

danielfrg:
yeah, i kinda decided to keep it in the "old" way of doing thing mostly for consistency with the python3 langpack.

Do you prefer to make the anaconda one use the new dev experience? I was thinking that when we update the regular python envs and templates we could also update the anaconda one.

I dont have strong opinions and we could start pushing that with the anaconda envs.

danielfrg:
I think i was also trying to make the anaconda one use the same as pipe as python3 so updating one would update the other but i dont think i like that idea anymore...

danielfrg:
OK, I updated it so we can start pushing the new dev experience :D

languages/anaconda3/config.json
```diff
@@ -1,5 +1,5 @@
 {
-    "display_name": "Python (Conda) Environments",
+    "display_name": "Conda Environments - beta",
```
platypii, line 2: Name seems inconsistent with others. Is there a standard here?

kennydaniel, line 2: Name seems inconsistent with others. Is there a standard here?

danielfrg: I keep coming back and forth with the name tbh, i think this is a bit better for the language selection since we could have more "Conda environments" that are not only Python. So something right now it would look like:

```
Conda environments:
 - Python 3.8 (Conda)
 - Python 3.9 (Conda)
 - Julia (Conda)
```

danielfrg: Updated to `Conda (Environments)` for consistency and the `(*)` gets removed from the UI.

danielfrg:
@zeryx or @kennydaniel could one of you do a final review? i made the changes we talked yesterday.

zeryx (approved):
Tested, LGTM

danielfrg:
Thanks!

PR #182, opened 2020-12-30
Duplicate Python 3.8 templates

danielfrg:
Duplicating this templates so that they work correctly based on the envpacks package-set name.


kennydaniel (approved):
:shipit:

PR #184, opened 2021-01-26
ALGO-915 minimum change for python security conflict [WIP]

zeryx:
https://algorithmia.atlassian.net/browse/ALGO-915
Updating langpack legacy images to avoid security conflicts discovered during customer call


dinoboy197:
@zeryx is this still needed? My guess is that nine months without a review means this isn't necessary or was worked around in some other way.

zeryx:
Agreed, thanks for catching this @dinoboy197 closing for now.

PR #187, opened 2021-02-13
[ALGO-934] Create Python 3.9 environment with updated base image

aslisabanci:
Couldn't test this on deep purple yet, a bit frustrated with the errors I've been getting so far. I'm opening this PR for your reviews, in case you notice something missing. Any help to test these is also appreciated to make things faster.

### DIFF
--- a/libraries/python39/Dockerfile
+++ b/libraries/python39/Dockerfile
@@ -55,11 +55,6 @@ RUN set -ex \
     \
     && wget -O python.tar.xz "https://www.python.org/ftp/python/${PYTHON_VERSION%%[a-z]*}/Python-$PYTHON_VERSION.tar.xz" \
     && wget -O python.tar.xz.asc "https://www.python.org/ftp/python/${PYTHON_VERSION%%[a-z]*}/Python-$PYTHON_VERSION.tar.xz.asc" \
-    # && export GNUPGHOME="$(mktemp -d)" \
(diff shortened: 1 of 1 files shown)


aslisabanci:
After commenting out the key auth, finally could test this with the following command on deep purple:
`./tools/environment_validator.py -g python3 -s python39 -t language -n python3`

libraries/python39/Dockerfile
```diff
@@ -0,0 +1,112 @@
...
+    && ln -s python3-config python-config
+
+# if this is called "PIP_VERSION", pip explodes with "ValueError: invalid truth value '<VERSION>'"
+ENV PYTHON_PIP_VERSION 21.0.1
+
+RUN set -ex; \
+    \
+    wget -O get-pip.py 'https://bootstrap.pypa.io/get-pip.py'; \
+    \
+    python get-pip.py \
```
kennydaniel: Depending on the base image, this might need to be `python3` in some cases? Might be fine as is though as long as we're controlling the base image.

kennydaniel (dismissed):

libraries/python39/Dockerfile
aslisabanci: Better be safe, I updated it to use `python3`

aslisabanci:
Python39+GPU langpack can be validated with the command, using CUDA10.2 as the base environment:
`./tools/environment_validator.py -b nvidia/cuda:10.2-cudnn7-runtime-ubuntu18.04 -g python3 -s python39 -t language -n python3 --nvidia-support 1`

Future selfs can refer to the conversation [here](https://algorithmia.slack.com/archives/G051R0WP2/p1613416562157700) for why we're using CUDA10.2 as our base img.

kennydaniel (approved):

PR #191, opened 2021-03-15
ALGO-943 initial python3-adk release

zeryx:
[ALGO-943](https://algorithmia.atlassian.net/browse/ALGO-904) -  Ship python ADK
This PR adds a new "language" for python (which preserves backwards compatibility with the old IPA system for python) that will provide a whole suite of new features and functionality. For more information please check out the ADK repo itself:
https://github.com/algorithmiaio/algorithmia-adk-python
Unit tests and integration tests have been added to the algorithmia-adk-python repo and CI gates release.

To test this, please run the following with the validator:
`./tools/environment_validator.py -g python3-adk -s python37 -t language -n python3-adk -c 1`
and test with `curl localhost:9999 -H 'Content-Type: application/json' -d '{"name": "Anthony"}'`
If that returns successfully, then the test has passed.

## Checklist
- [x] My PR title includes a relevant Jira ticket name
- [x] If I made configuration changes, they are in the config template in the deploy directory
- [x] I have added unit tests where appropriate
- [x] I have added integration tests where appropriate
- [ ] Manual tests are required in a cluster


languages/python3-adk/template/algorithmia.conf
```diff
@@ -0,0 +1,5 @@
+{
+    "username": "__USER__",
+    "algoname": "__ALGO__",
+    "langauge": "python3.7"
```
aslisabanci: I'm not clear in how we use these .conf files, but in case they're functional and we want to keep them around; we can fix the typo of `langauge`->`language`

aslisabanci (dismissed):

languages/python3-adk/template/algorithmia.conf
zeryx: yeah good point, this might be a good question for @algorithmiaio/product-services as I know the concept of `language` is defined in many locations. I believe here it's just user facing information, but that might be worth adjusting

zeryx: good news, that language field is actually just used in the template repo now, it's completely detatched from the rest of the system, so we can set that to be `python3.x-adk` or.. anything really

aslisabanci (dismissed):

languages/python3-adk/template/src/Algorithm.py
```diff
@@ -0,0 +1,10 @@
+from Algorithmia import ADK
+
+# API calls will begin at the apply() method, with the request body passed as 'input'
+# For more details, see algorithmia.com/developers/algorithm-development/languages
+def apply(input):
```
kennydaniel: Why doesn't the ADK template have a load function?

kennydaniel:
Overall looks good with some minor comments. The load function missing was a big part of the goal with ADK i thought? Also the name is a bit long "Python 3.x (Environments) - Algorithm Development Kit". I think languages don't generally include "(Environments)"?

We can merge in langpacks, but we should hold deploying the environment until docs are ready. The ADK readme is solid, I think that mostly would need to be added to devcenter.

languages/python3-adk/template/src/Algorithm.py
zeryx: Did we want to have the template as simple as possible, or did we want the default to have a load step? I'm down with having a load step be the default, but as users can't choose which template they want to see it is slightly more advanced than the default.

kennydaniel (approved):

PR #192, opened 2021-03-16
ALGO-965 Spacy 3.0.x support

zeryx:
[ALGO-965](https://algorithmia.atlassian.net/browse/ALGO-965) - Spacy 3.0.x Support

This PR adds support for Spacy 3.0.x, which can be verified with the following validator command:
`./tools/environment_validator.py -b ubuntu:20.04 -g python3 -s python37 -d spacy-3.0.x -t dependency -n spacy-3.0.x`

### DIFF
--- a/libraries/spacy-3.0.x/Dockerfile
+++ b/libraries/spacy-3.0.x/Dockerfile
@@ -1,0 +1,9 @@
+
+RUN pip install "spacy>=3.0,<3.1"
+RUN python -m spacy download en_core_web_sm
+RUN python -m spacy download de_core_news_sm
+RUN python -m spacy download es_core_news_sm
+RUN python -m spacy download pt_core_news_sm
+RUN python -m spacy download fr_core_news_sm
+RUN python -m spacy download it_core_news_sm
+RUN python -m spacy download nl_core_news_sm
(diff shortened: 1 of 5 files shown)


libraries/spacy-3.0.x/Dockerfile
```diff
@@ -0,0 +1,9 @@
+
+RUN pip install "spacy>=3.0,<3.1"
+RUN python -m spacy download en_core_web_sm
+RUN python -m spacy download de_core_news_sm
+RUN python -m spacy download es_core_news_sm
+RUN python -m spacy download pt_core_news_sm
+RUN python -m spacy download fr_core_news_sm
+RUN python -m spacy download it_core_news_sm
+RUN python -m spacy download nl_core_news_sm
```
kennydaniel, line 9: How large are these languages?

zeryx: > How large are these languages?

Looks like the largest is 22 MB, and the smallest is 12 MB; overall very lightweight, which is great! Unfortunately not super easy to install programmatically from within a spacy script as they really like their CLI :joy:

kennydaniel (approved):

PR #206, opened 2021-07-08
updated new mechanism for verifying adk support (.adk dot file)

zeryx:
As per our conversation @kennydaniel , this mechanism preserves the existing naming convention of python algorithms, while still being able to verify and validate if an environment is "new" by an `.adk` dotfile.

To test, run the following two tests:
`./tools/environment_validator.py -g python3 -s python37 -t language -n python3 -a 1` if it succeeds, then an algorithm with the old naming convention works.
`./tools/environment_validator.py -g python3 -s python37 -t dependency -n test_python3_legacy`  if that continues and passing in default input succeeds; all tests pass.


danielfrg (approved):
If i remember correctly akd is its own python pkg right? if thats the case we should add it to the requirements.txt, if not and its included in the client lib nevermind :)

zeryx:
> 
> 
> If i remember correctly akd is its own python pkg right? if thats the case we should add it to the requirements.txt, if not and its included in the client lib nevermind :)

It's in the client! It's distinct but we have it as a dependency to algorithmia-python here
https://github.com/algorithmiaio/algorithmia-python/blob/develop/requirements.txt#L7
https://github.com/algorithmiaio/algorithmia-adk-python is the main repo

PR #213, opened 2021-11-12
Creating a DataRobot Java CodeGen Template

zeryx:
[AML-10](https://datarobot.atlassian.net/browse/AML-10) - Datarobot Java CodeGen

We're creating a simple template that utilizes a publicly available model file that uses.

## Checklist
- [x] My PR title includes a relevant Jira ticket name
- [x] If I made configuration changes, they are in the config template in the deploy directory
- [x] I have added unit tests where appropriate
- [x] I have added integration tests where appropriate
- [x] Manual tests are required in a cluster


templates/datarobot-java/model_manifest.json.freeze
```diff
@@ -0,0 +1 @@
+{"required_files": [{"name": "sentiment_model", "source_uri": "data://demo/datarobot_models/618e9e830d2eca5ba9cd1014.jar", "md5_checksum": "6ebc07d2f1f0edffef1bc6aff24c7b93"}], "optional_files": [], "timestamp": "1636737821.9979346", "lock_checksum": "80d2a2056950f038eea78c8bb6f8c8a2"}
```
kennydaniel, line 1: I'm not sure I like having a pre-frozen file in the template

aslisabanci (commented):
I have a comment from a fundamental point: `py4j` is not the only way to execute codegen models. If the downloaded codegen jar is the model+monitoring jar, the algorithm coder can just get their predictions by running the jar with certain arguments, like: 
```
java -jar <local path to scoring code jar> csv \
  --input=<local path to input CSV> \
  --output=<local path to output CSV> \
  --enable_mlops \
  --dr_token=<your api token> 
```
This approach doesn't require importing `py4j` or the template code here. 
So the algo code will be like:
```
from Algorithmia import ADK
import Algorithmia
import csv
import subprocess
import os
import time


def load(modelData):
    modelData.user_data["model_path"] = modelData.get_model("datarobot_codegen_with_monitoring")
    modelData.system_data["dr_token"] = os.environ["DATAROBOT_API_TOKEN"]
    return modelData


def apply(input, modelData):
    model_path = modelData.user_data["model_path"]
    dr_token = modelData.system_data["dr_token"]
    
    input_csv = modelData.client.file(input).getFile().name

    start = time.time()
    cmd = f'java -jar {model_path} ' \
          f'csv --input={input_csv} ' \
          '--output=- --enable_mlops ' \
          f'--dr_token="{dr_token}"'
    p = subprocess.check_output(cmd, shell=True)
    pred_duration = round(time.time() - start)
    return {"predictions": p.decode('ascii'), 'prediction_duration_seconds': pred_duration}

client = Algorithmia.client()
algorithm = ADK(apply, load, client)
algorithm.init()
```
> 

In summary, not everybody will run a codegen model using the templated code or the requirements. What do you think about this distinction? @zeryx @kennydaniel

templates/datarobot-java/model_manifest.json.freeze
zeryx: it also links to a model file that won't necessarily move with the environment (at least not automatically) between clusters. Not sure the best path here but the DR Template doesn't do much without a model

zeryx:
@aslisabanci I think both approaches can work, and the subprocess can be simpler - but it runs into performance issues as it shuts down and starts back up the JVM, and requires reloading the model on each call. This has been something most of our DR customers have asked for is improved performance, which is why we utilise the py4j system. If there's an easier way to manage the apply function and simplify the IO that would be awesome

templates/datarobot-java/model_manifest.json.freeze
zeryx: Without a frozen template this algorithm wouldn't be able to compile using the web editor or run without the `algo freeze`. Alternatively we could add an `algo freeze` operation to the build process via the web editor

aslisabanci: How about modifying the ADK to not require a ".freeze" file? The user may have a model_manifest.json in their algo repo, but maybe they haven't frozen this yet? They may be still experimenting with the algo and unless the algo is ready to be published, a frozen manifest might not be needed in practice.
If there's a .freeze file though, we can infer that this user wants to protect the model and in that case ADK can check the hashes, etc.

aslisabanci:
> @aslisabanci I think both approaches can work, and the subprocess can be simpler - but it runs into performance issues as it shuts down and starts back up the JVM, and requires reloading the model on each call. This has been something most of our DR customers have asked for is improved performance, which is why we utilise the py4j system. If there's an easier way to manage the apply function and simplify the IO that would be awesome

I don't object to the perf aspect, but sometimes the user won't know how to use the "codegen + monitoring" jar using py4j. This jar is invoked with certain parameters and since we don't know the internals of this Java package's implementation, so we won't know how to write the py4j wrapper. Taking my example in this thread, I don't know how I could call this jar package with these parameters using py4j:
```
java -jar <local path to scoring code jar> csv \
  --input=<local path to input CSV> \
  --output=<local path to output CSV> \
  --enable_mlops \
  --dr_token=<your api token>
  ```
So for me, it wasn't a matter of preference, but a necessity to call this jar using subprocess.

We can keep the template like this if it fits the majority of our interested users' way of using the codegens, but I wanted us to be aware of the other use cases.

templates/datarobot-java/model_manifest.json.freeze
zeryx: that's an interesting idea, if there's no freeze file we just download the .json data and don't check for version conflicts. I think that's a quite reasonable PR and a healthy "failover" solution

